PR #169, opened 2019-10-30
Buildstage build() method - fix misspelt Jenkinsfile

sujikann:


kmanning:
See: https://github.com/manheim/terraform-pipeline/tree/v5.2

Let's add a unit test, just for the sake of exercising the method and catching these sorts of typos in the future.

kmanning:
Fixed in https://github.com/manheim/terraform-pipeline/pull/170 - Thanks for the find and issue report!

Issue #109, opened 2019-09-03
Terraform 0.12 Compatibility


kmanning:
Update and support Terraform 0.12. For information on the breaking changes to plugins see:

https://www.terraform.io/docs/extend/terraform-0.12-compatibility.html

https://github.com/hashicorp/terraform/issues/21454

TF 0.12: `terraform validate -check-variables` flag not defined

missing required provider arguments should be a WARNING not an ERROR

https://github.com/hashicorp/terraform/issues/21408

kmanning:
Work done so far by the PR above:
1. As of terraform 0.12 and above, `terraform validate` now needs a `terraform init`.
2. The PR above solves this problem by  adding `terraform init -backend=false` to the TerraformValidateStage.
3.  `-backend=false` is required for TerraformValidateStage, only because many/most of the backend plugins are environment-based, and TerraformValidateStage was never designed to have environment configuration. 

Work left to do:
1. The before/after hooks  in TerraformValidateStage are insufficient.  You can  only decorate around the entirety of `delete/checkout/validate`.  Offer more flexibility here, to allow decorations around  `terraform validate` itself.
2.  Once that's available, remove the hardcoded `terraform init` introduced by the PR above.
3.  Modify TerraformPlugin to conditionally inject `terraform init` for 0.12+.  For versions prior  to 0.12, leave the functionality as-is prior to the first PR for this issue (that is, prior to 0.12, TerraformValidate should NOT run `terraform init -backend=false`.

kmanning:
Released in v5.3 https://github.com/manheim/terraform-pipeline/releases/tag/v5.3

PR #173, opened 2019-12-12
Fix issue #109 (Terraform 0.12 Support)

dkolb:
~Per the changes in the README, this adds a specific method call, `withVersion12Support()` to the `TerraformValidateStage` class.  Internally, the `TerraformValidateCommand.instance()` now returns a command with no arguments while `TerraformValidateCommand.instanceWithCheckVars()` holds the previous behavior.~

Creates a `TerraformPlugin` that is a default plugin and detects `.terraform-version` files.  Also exposes a `withTerraformVersion(String)` method in the event you cannot use the dot file method.  The plugin compares semantic versions to determine behaviors that need modification on various Commands and Stages.  Currently only worries itself with the validate Stage and Command as that is where breaking changes were made in 0.12.

Furthermore, the `TerraformInitCommand` now has a default constructor that assigns an empty string to the environment.  You can call `withoutBackend()`, which will always set the `-backend=false` argument and ignore any `-backend-config=` arguments. The empty string is used because some plugins (`S3BackendPlugin` for example) assume that the environment is not null.  Who knows what external plugins assume that also.

The `-backend=false` option is supported on the `terraform init` command as far back as `0.9.0`.  This holds with the current implied supported versions, since `0.8.x` requires a SOURCE argument and does not support executing `terraform init [-dash-args] ...]` in a directly.

TL;DR: All changes are supported from 0.9.0 and up, which is already what we supported.  To enable >=0.12.0 support properly you need to set a flag on the Validate stage.

### DIFF
--- a/src/TerraformInitCommand.groovy
+++ b/src/TerraformInitCommand.groovy
@@ -1,14 +1,21 @@
 class TerraformInitCommand {
+    private static final DEFAULT_PLUGINS = []
     private boolean input = false
     private String terraformBinary = "terraform"
     private String command = "init"
     String environment
     private prefixes = []
     private backendConfigs = []
+    private boolean doBackend = true
     private String directory
 
-    private static plugins = []
+    private static plugins = DEFAULT_PLUGINS.clone()
     private appliedPlugins = []
+
+    private TerraformInitCommand() {
+        //We need this because some plugins expect a non-null value here.
+        this('')
+    }
 
     public TerraformInitCommand(String environment) {
         this.environment = environment
@@ -34,6 +41,11 @@ class TerraformInitCommand {
         return this
     }
 
+    public TerraformInitCommand withoutBackend() {
+        this.doBackend = false
+        return this
+    }
+
     public String toString() {
         applyPluginsOnce()
 
@@ -44,8 +56,12 @@ class TerraformInitCommand {
         if (!input) {
             pieces << "-input=false"
         }
-        backendConfigs.each { config ->
-            pieces << "-backend-config=${config}"
+        if(doBackend) {
+            backendConfigs.each { config ->
+                pieces << "-backend-config=${config}"
+            }
+        } else {
+            pieces << "-backend=false"
         }
         if (directory) {
             pieces << directory
@@ -71,12 +87,16 @@ class TerraformInitCommand {
         return new TerraformInitCommand(environment)
     }
 
+    public static TerraformInitCommand instance() {
+        return new TerraformInitCommand().withoutBackend()
+    }
+
     public static getPlugins() {
         return plugins
     }
 
     public static resetPlugins() {
-        this.plugins = []
+        this.plugins = DEFAULT_PLUGINS.clone()
         // This is awkward - what about the applied plugins...?
     }
 }
--- a/src/TerraformInitCommand.groovy
+++ b/src/TerraformInitCommand.groovy
@@ -12,7 +12,7 @@ class TerraformInitCommand {
     private static plugins = DEFAULT_PLUGINS.clone()
     private appliedPlugins = []
 
-    private TerraformInitCommand() {
+    TerraformInitCommand() {
         //We need this because some plugins expect a non-null value here.
         this('')
     }
--- a/src/TerraformValidateStage.groovy
+++ b/src/TerraformValidateStage.groovy
@@ -1,10 +1,9 @@
 class TerraformValidateStage implements Stage {
-
+    private static final DEFAULT_PLUGINS = []
     private Jenkinsfile jenkinsfile
-    private TerraformValidateCommand validateCommand
     private Map<String,Closure> decorations
 
-    private static plugins = []
+    private static plugins = DEFAULT_PLUGINS.clone()
 
     public static final String ALL = 'all'
 
@@ -26,12 +25,15 @@ class TerraformValidateStage implements Stage {
 
         def validateCommand = TerraformValidateCommand.instance()
 
+        def initCommand = TerraformInitCommand.instance()
+
         return {
             node {
                 applyDecorations(ALL) {
                     stage("validate") {
                         deleteDir()
                         checkout(scm)
+                        sh initCommand.toString()
                         sh validateCommand.toString()
                     }
                 }
@@ -84,6 +86,6 @@ class TerraformValidateStage implements Stage {
     }
 
     public static void resetPlugins() {
-        this.plugins = []
+        this.plugins = DEFAULT_PLUGINS.clone()
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Added the [TfvarsFilesPlugin](./docs/TfvarsFilesPlugin.md) to support adding var files to plan and apply commands.
+
 # v5.2
 
 * [Issue #168](https://github.com/manheim/terraform-pipeline/issues/168) Fix typo in Jenkinsfile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Added the [TfvarsFilesPlugin](./docs/TfvarsFilesPlugin.md) to support adding var files to plan and apply commands.
+* [Issue #109](https://github.com/manheim/terraform-pipeline/issues/109) Support Terraform 0.12
 
 # v5.2
 
--- a/README.md
+++ b/README.md
@@ -2,15 +2,30 @@
 A reusable pipeline template to apply terraform configuration serially across multiple environments.
 
 # Requirements and Assumptions
-1.  You have a separate terraform project that can be run with the following commands:
+1.  You are using at least terraform 0.9.x (must support `-backend=false`)
+2.  You have a separate terraform project that can be run with the following commands:
     1.  `terraform init`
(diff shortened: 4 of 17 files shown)


kmanning:
Can you pull against v5.3-dev branch?

dkolb:
Done.

dkolb:
Note that this does nothing to address hashicorp/terraform#21408.  While they have accepted it as a bug, I don't see much progress on it.  The workarounds are to either set the region in the aws provider to a terraform variable or set the `AWS_DEFAULT_REGION` environment variable to a placeholder value when running validate.

kmanning:
Can you help me understand the `-backend=false` piece? I’m not entirely caught up on 0.12 changes.  

1) My understanding is that you only need `-backend=false` to disable a previously initialized backend https://www.terraform.io/docs/commands/init.html#backend-initialization.  By default,  TerraformEnvironmentStage will not initialize a backend (assuming you haven’t enabled a plugin that will add a backendConfig. As such, `backend=false` shouldn’t be necessary - but as it’s written, it will  always be  added unless a backConfig is added. That doesn’t seem like a necessity.  If what I’ve said is correct, I think I’d prefer to change `initWithoutBackend` to  instead be `withoutBackend`, which could then simply append `backend=false` to backendConfigs.  <— this allows you to explicitly undo a previously initialized backend, if that’s what you explicitly  want to do,  but  if NOT, then the previous behavior remains (meaning, if no backends were configured, then backend=false will not be set).

2) Can we be more clever with how we handle versions?  `withVersion12Support` sets an unpleasant precedence - what  happens if version 0.13 creates breaking changes?  Do we add a  `withVersion13Support`?  Ack.  What this issue reveals is a poor choice on my part - https://github.com/manheim/terraform-pipeline/blob/0233f1e41ab83d442e7f8736b41621b19ca81750/src/TerraformValidateCommand.groovy#L62.  I should never have REQUIRED that `-check-variables=false` be part of the TerraformValidateCommand. <-- During  the initial development of this library, I made several of these poor choices.  What I did later, was to break this behavior out into "default plugins", and then later expand on the plugin's behavior. 

If I were to try to take that approach for this issue, I think maybe we need a `TerraformPlugin`.  Maybe if such a Plugin existed, it could have  a `withVersion()` method, we could then isolate all the version-specific behavior in one place.

* `TerraformPlugin.init()` - would implement the existing behavior across the appropriate plugins (using pre-0.12 flags like -check-variables)
* `TerraformPlugin.withVersion('0.12').init()` - would *NOT*  add `check-variables`
* `TerraformPlugin.withVersion('FutureBreakingVersion').init()` - would do other things that specific to that future breaking version.

^-- We could even be clever, and take the solution above, and skip and explicit call  to `withVersion`.  What if we checked for the presence of a `.terraform-version` file?  If such a file existed, we could automatically call `withVersion` with the appropriate value.  Then terraform-pipeline would auto-configure for the appropriate version without the user needing to do anything at all.

src/TerraformInitCommand.groovy
```diff
@@ -44,8 +55,12 @@ class TerraformInitCommand {
...
             pieces << "-input=false"
         }
-        backendConfigs.each { config ->
-            pieces << "-backend-config=${config}"
+        if(doBackend) {
+            backendConfigs.each { config ->
+                pieces << "-backend-config=${config}"
+            }
+        } else {
+            pieces << "-backend=false"
```
kmanning: See comment

src/TerraformValidateStage.groovy
```diff
@@ -13,6 +14,11 @@ class TerraformValidateStage implements Stage {
         this.decorations = new HashMap<String,Closure>()
     }
 
+    public TerraformValidateStage withVersion12Support() {
```
kmanning: 2) Can we be more clever with how we handle versions?  `withVersion12Support` sets an unpleasant precedence - what  happens if version 0.13 creates breaking changes?  Do we add a  `withVersion13Support`?  Ack.  What this issue reveals is a poor choice on my part - https://github.com/manheim/terraform-pipeline/blob/0233f1e41ab83d442e7f8736b41621b19ca81750/src/TerraformValidateCommand.groovy#L62.  I should never have REQUIRED that `-check-variables=false` be part of the TerraformValidateCommand. <-- During  the initial development of this library, I made several of these poor choices.  What I did later, was to break this behavior out into "default plugins", and then later expand on the plugin's behavior. 

If I were to try to take that approach for this issue, I think maybe we need a `TerraformPlugin`.  Maybe if such a Plugin existed, it could have  a `withVersion()` method, we could then isolate all the version-specific behavior in one place.

* `TerraformPlugin.init()` - would implement the existing behavior across the appropriate plugins (using pre-0.12 flags like -check-variables)
* `TerraformPlugin.withVersion('0.12').init()` - would *NOT*  add `check-variables`
* `TerraformPlugin.withVersion('FutureBreakingVersion').init()` - would do other things that specific to that future breaking version.

^-- We could even be clever, and take the solution above, and skip and explicit call  to `withVersion`.  What if we checked for the presence of a `.terraform-version` file?  If such a file existed, we could automatically call `withVersion` with the appropriate value.  Then terraform-pipeline would auto-configure for the appropriate version without the user needing to do anything at all.

src/TerraformInitCommand.groovy
dkolb: Given a provider configuration of `terraform { backend { s3 { } }`, a plain `terraform init -input=false` will fail claiming that you have not provided a region, bucket, or key.  Since almost all the backend configs are environment dependent, and the validate stage is effectively without an environment, `-backend=false` is required to _skip_ the config and avoid these errors.  backendConfigs has to be skipped because your backend plugins are going still write those arguments to that array.

We aren't undoing a previously initialized backend, we are skipping the backend.

> To skip backend configuration, use `-backend=false`. Note that some other init steps require an initialized backend, so it is recommended to use this flag only when the working directory was already previously initialized for a particular backend.

I'll work on reworking it to `withoutBackend()` pattern.

src/TerraformValidateStage.groovy
dkolb: I have no problem looking for a `.terraform-version` file, I use tfenv myself.  I just didn't want to inflict the pattern on anyone else.

I believe a default plugin might work and could handle both `withVersion(String)` and detecting from the `.terraform-version`.

dkolb:
I just want to say for the record I rebased on the updated v5.3-dev branch and it still gave me those merge conflicts. -_-

Anyway, rewritten as a plugin.  Squashed.  Successful run using the code here:

https://account-management-bento.awsmancsnp.manheim.com/blue/organizations/jenkins/PingFederate%2Fmanheim-pingfederate-docker/detail/master/62/pipeline

The failure is because there we haven't added the `e2e.tfvars` file and is unrelated to the terraform 0.12 support.

src/TerraformInitCommand.groovy
kmanning: I read your answer, but was previously still not fully understanding.  I've since installed 0.12.8, and played around with it, and better understand your change now.  The piece that I was missing was:

1) Prior to 0.12.*, you could run `terraform validate` without first running `terraform init`
2) As of 0.12.+, you are forced to run `terraform init` prior to being able to run `terraform validate`

I'm mulling over the change that you're making - adding `terraform init` to the TerraformValidateStage, then setting `-backend=false`.   In an ideal world, I would push the addition of `terraform init` into TerraformPlugin as well - that is, prior to 0.12, `terraform init` would not be called as part of TerraformValidateStage, but 0.12+ would have the command added.  

TerraformValidateStage is not currently that flexible.

I'm not ignoring this PR, just mulling over the consequences.  I'd be willing to accept this PR into the v5.3-dev branch, but I think I'd want to continue building on your PR before publishing an official release.

dkolb: I get it.  In the long term, you want the plugin to be the source of truth for how things need to be adjusted between versions.  Currently the only real way without a deep rework of Stages and their interface the best we can do is keep adding switches and have the plugin flip those switches.  Which while better than the original implementation, still spreads out the logic of 0.11 vs 0.12 vs future versions across a lot of different classes.

I've got a fork I'm running off of in the meantime, so take your time.  It's a needly problem. especially with the way the terraform devs seem willing to completely rework the workflow for various ... goals so to speak, as 0.12 showed.

Outside of adding `runBefore()` and `runAfter()` stages to every single command, what happens in the future when 1.0 decides validate is just "plan" with "-check-variables=false -refresh-state=false"?!?  It seems like we are approaching a world where the plugin would need to pick and register a Stage Implementation and at that point are we over engineering a solution to a problem that hasn't yet occured?

Having typed all this out, I think pre and post hooks on the Stages might just be the best answer until a terraform version throws us a new wrench that doesn't solve.

kmanning: ^---  Yeah, pre/post hooks were exactly what I had in mind.  A pre/post hook around JUST `terraform validate`, would give `TerraformPlugin` the ability to conditionally inject a `terraform init` based on the version being used.

I'll take one last look at the changes, merge this change in, note this discussion on Issue #109 as remaining work before closing it out, then I'll attack it with a PR when  I get the chance.

Thank you again!

kmanning:
Merging PR.  The Issue #109 was updated with information about more work to be done before publishing an official release.

kmanning:
Continued in https://github.com/manheim/terraform-pipeline/pull/177

PR #188, opened 2020-02-06
Issue 87 Part 3: Add codenarc and fix terraform-version comparison

kmanning:
* The initial merge of issue_87_part_3 introduced a regression, when refactoring version comparison to use `>=`.
* The compareTo method of SemanticVersion doesn't properly override the various comparison operators.  It also behaves differently in unit tests than it does under a normal Jenkinsfile context.
* Switch to `<=>` operator instead of `>=` when comparing semantic versions, to determine whether or not to use Terraform v0.12 workflow.

* Test: `new SemanticVersion('0.12.0') >= new SemanticVersion('0.12.0')`
* Expected: `true`
* Actual: `0`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,6 @@
 apply plugin: 'groovy'
 apply plugin: 'jacoco'
+apply plugin: 'codenarc'
 
 repositories {
     mavenCentral()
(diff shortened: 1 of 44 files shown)


kmanning:
Note: this problem was identified in the original author's PR - https://github.com/manheim/terraform-pipeline/pull/173/files#diff-2724a09cdd048285c15140f07b77fb5dR47-R51

The return of `int` instead of `boolean` appears to be due to CPS .

PR #197, opened 2020-03-06
fix error with not using validate stage

bondezbond:
groovy.lang.MissingPropertyException: No such property: nextStage for class: TerraformEnvironmentStage


kmanning:
See Issue  #198

kmanning:
Thanks for the fix!

- Could you add a unit test?  Just exercising the method should show the failure.

PR #220, opened 2020-05-11
Update README to discuss plugin intialization order.

### LINKED ISSUE #206: Plugin Initialization Order Effects are Uncodumented
## Update

Updating this issue to a documentation issue and will open a PR Soon ™️ with updates to the relevant sections of README.md.   Essentially we should just document that plugins are initialized in an "inside out" order where the last plugin initialized will have it's decorations wrapping all other plugin decorations.  Specifically we should call out that `TerraformEnvironmentStage.withGlobalEnv(String, String)` should be the last thing you do so that your environment variables are available to all other variables.

## Original Issue

```groovy
// Jenkinsfile
@Library(['terraform-pipeline-modern', 'pingfederate-pipeline-customizations']) _

Jenkinsfile.init(this)
Jenkinsfile.declarative = true
Jenkinsfile.pipelineTemplate = this.PingFederatePipelineTemplateLowers
Jenkinsfile.defaultNodeName = 'tfenv'

TerraformEnvironmentStage.withGlobalEnv('AWS_ROLE_ARN', 'role::arn::here')

WithAwsPlugin.init()
```

This will result in withAWS being skipped due to the withEnv from the EnvironmentVariablePlugin's decorations being *after* withGlobalEnv.

Workarounds are to directly manipulate `env` at the top of your file:
```groovy
env.PERF_AWS_ROLE_ARN = env.PROD_AWS_ROLE_ARN
```

I'm not sure if we should fix this or just update documentation on the withAWS plugin page.

dkolb:
Closes #206 by updating docs to basically point out it's a feature kthnksbye.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,55 @@ validate.then(deployQa)
 
 The example above gives you a bare-bones pipeline, and there may be Jenkinsfile features that you'd like to take advantage of.  Some of these features have been pre-defined as Plugins for this library.  Pre-defined plugins can be enabled by simply calling their static `init()` method.
 
+Plugins primarily function by wrapping Jenkins DSL in blocks, such as `withEnv() {}` or `withAWS() {}`. This means that they nest inside each other. The order you initialize the plugins is maintained and translates to an inside to outside order.
+
+Of primary concern is using `TerraformEnvironmentStage.withGlobalEnv(String, String)` to define environment variables used by plugins. This convienence method actually instantiates an instance of the `EnvironmentVariablePlugin` under the covers.
+
+Take the following example:
+
+```
+TerraformEnvironmentStage.withGlobalEnv('AWS_ROLE_ARN', 'role::arn::here')
+
+WithAwsPlugin.init()
+```
+
+This will actually end up with a Jenkins Pipeline that looks like this:
+
+```
+{
+  /* Code that looks for iam role in env vars. */
+  if(iamRole != null) {
+    withAws(role: iamRole) {
+      withEnv("AWS_ROLE_ARN=role::arn::here") {
+        //Terraform commands
+      }
+    }
+  } else {
+    sh "echo no role found skipping withAWS)"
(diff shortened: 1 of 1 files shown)


kmanning:
I like core of the issue that you raise - ordering of plugins is important, and under the User's control.

I'd prefer to document this in a more generic way.

To be more specific:
```
Plugins primarily function by wrapping Jenkins DSL in blocks, such as `withEnv() {}` or `withAWS() {}`. This means that they nest inside each other. The order you initialize the plugins is maintained and translates to an inside to outside order.
```
^-- I like that.  It's clear.

```
Of primary concern is using `TerraformEnvironmentStage.withGlobalEnv(String, String)` to define environment variables used by plugins. This convienence method actually instantiates an instance of the `EnvironmentVariablePlugin` under the covers.
```
^-- that's starting to get confusing.  And it's confusing because the library itself is confusing, and it's due to the fact that `withGlobalEnv` is  a plugin is surprising.

Rather than pick `withGlobalEnv` and `EnvironmentVariablePlugin`, can we pick 2 plugins that illustrate your problem, and bypass the distinct issue with EnvironmentVariablePlugin?  EnvironmentVariablePlugin was an attempt to start addressing the underlying problem - but it's still not good enough (as evidenced by the problems you're having).  I'd rather avoid documenting that specific problem (because I'll be more inclined to not fix the problem, and people will be more inclined to get used to the problem), and instead, express the desired outcome.  

In this case, the desired outcome is, "Jenkinsfile DSL ordering matter, and therefore, the order of terraform-pipeline plugins matter.  Just as you control the Jenkinsfile DSL  ordering to accomplish what you need, you control the order that you initialize plugins to accomplish what you need."

dkolb:
Hmmm.  Yes I can see that being a problem.  Let me think on that and rummage around and see if I have a better example.  I'm open to any internal examples you may have (send them via Slack) and I'll adapt them to the documentation.

dkolb:
Closing in favor of #239

Issue #193, opened 2020-02-18
Support passing branch plans back to Github PR


bondezbond:
If I use multi-branch pipeline with PRs from github. Would like to pass back the pr `plan` output back to the github pr

kmanning:
Simplify the current implementation.  

* Repo slug can be constructed from the scm URL
* The base URL can be constructed from the scm URL
* Remove landscape functionality from the plugin - there's a separate Landscape plugin

PR #228, opened 2020-06-08
Support PR plan output to github comments

jleopold28:
Implementation for https://github.com/manheim/terraform-pipeline/issues/193

Added new plugin `TerraformPlanResultsPR` which will comment Terraform plan results as github PR comments.

### DIFF
--- a/docs/TerraformPlanResultsPR.md
+++ b/docs/TerraformPlanResultsPR.md
@@ -8,6 +8,7 @@ One-Time Setup:
 * Install the terraform-landscape gem on your Jenkins slaves. (ONLY when using `withLandscape(true)`)
 
 Requirements:
+* Enable CredentialsPlugin to set the GITHUB_TOKEN environment varaible. This is later used in the TerraformPlanResultsPR Plugin for authentication.
 * Enable AnsiColorPlugin for colors in Jenkins (ONLY when usng `withLandscape(true)`)
 
 ```
@@ -15,9 +16,14 @@ Requirements:
 
 Jenkinsfile.init(this)
 
+// FOO/GITHUB_TOKEN will contain the respective username/password values of the 'my-cred' credential.
+CredentialsPlugin.withBuildCredentials([usernameVariable: 'FOO', passwordVariable: 'GITHUB_TOKEN'], 'my-cred').init()
+
 AnsiColorPlugin.init()                                                               // REQUIRED: Decorate your TerraformEnvironmentStages with the AnsiColor plugin
-TerraformPlanResultsPR.withLandscape(true).withRepoSlug('reposlug').init()           // Post the plan results in the comments of a PR (landscape_gem used = true)
-// OR TerraformPlanResultsPR.withLandscape(false).withRepoSlug('reposlug').init()    // Post the plan results in the comments of a PR (landscape_gem used = false)
+TerraformPlanResultsPR.withRepoHost("https://github.com/api/v3/")
+                      .withRepoSlug("my-org/my-repo")
+                      .withLandscape(true)
+                      .init()
 
 def validate = new TerraformValidateStage()
 def deployQa = new TerraformEnvironmentStage('qa')
--- a/docs/TerraformPlanResultsPR.md
+++ b/docs/TerraformPlanResultsPR.md
@@ -2,8 +2,6 @@
 
 Use this to post Terraform plan results in the comments of a PR.
 
-Configure usage of terraform landscape with the `withLandscape(bool)` method.
-
 One-Time Setup:
 * Install the terraform-landscape gem on your Jenkins slaves. (ONLY when using `withLandscape(true)`)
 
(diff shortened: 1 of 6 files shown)


kmanning:
Overall - nice job.

Now that I'm seeing the finished product, it's more complex/involved than I realized.  It's forcing me to ask the question, "Does this belong as a core plugin for terraform-pipeline?  Or would this feature  be more appropriate in a separate library?"

While the feature is useful, I'm struggling to figure out whether or not it will be useful outside of CAI?  The complexity in particular feels like a source of a lot of potential future churn, that may be better isolated in a separate project?

I'm open to discussion on the above topic.

docs/TerraformPlanResultsPR.md
```diff
@@ -0,0 +1,47 @@
+## [TerraformPlanResultsPR](../src/TerraformPlanResultsPR.groovy)
+
+Use this to post Terraform plan results in the comments of a PR.
+
+One-Time Setup:
+* Install the terraform-landscape gem on your Jenkins slaves. (ONLY when using `withLandscape(true)`)
+
+Requirements:
+* Enable CredentialsPlugin to set the GITHUB_TOKEN environment varaible. This is later used in the TerraformPlanResultsPR Plugin for authentication.
```
kmanning, line 9: typo: "varaible"

docs/TerraformPlanResultsPR.md
```diff
@@ -0,0 +1,47 @@
...
+```
+@Library(['terraform-pipeline']) _
+
+Jenkinsfile.init(this)
+
+// Required to set GITHUB_TOKEN envionrment variable for use with TerraformPlanResultsPR Plugin
+// FOO/GITHUB_TOKEN will contain the respective username/password values of the 'my-cred' credential.
+CredentialsPlugin.withBuildCredentials([usernameVariable: 'FOO', passwordVariable: 'GITHUB_TOKEN'], 'my-cred').init()
+
+AnsiColorPlugin.init()                                           // Required when using 'withLandscape(true)'
```
kmanning, line 32: What does 'Required  when using `withLandscape` mean?  What happens if I enable landscape and do not enable  the AnsiColorPlugin?

docs/TerraformPlanResultsPR.md
kmanning: Separately: I'm inclined to avoid saying specifically that  the CredentialsPlugin needs to be used.  You've written a plugin that accepts *any* environment variable, so I'd be inclined to let the User decide how that environment variable is created (with the intent of reducing dependencies between plugins).

jleopold28: * typo has been fixed
* Updated docs and removed requirement for CredentialsPlugin

docs/TerraformPlanResultsPR.md
jleopold28: I have updated the docs to expand on this. When using the Landscape gem,AnsiColor is required for Jenkins to colorize the output.

jleopold28: If you enable landscape without AnsiColor, you will see the plan results from `terraform plan | landscape` without color

docs/TerraformPlanResultsPR.md
```diff
@@ -0,0 +1,47 @@
+## [TerraformPlanResultsPR](../src/TerraformPlanResultsPR.groovy)
+
+Use this to post Terraform plan results in the comments of a PR.
+
+One-Time Setup:
+* Install the terraform-landscape gem on your Jenkins slaves. (ONLY when using `withLandscape(true)`)
```
jantman, line 6: Can we please change "Jenkins slaves" to "Jenkins agents" or "Jenkins build nodes"?

jantman:
> Overall - nice job.
> 
> Now that I'm seeing the finished product, it's more complex/involved than I realized. It's forcing me to ask the question, "Does this belong as a core plugin for terraform-pipeline? Or would this feature be more appropriate in a separate library?"
> 
> While the feature is useful, I'm struggling to figure out whether or not it will be useful outside of CAI? The complexity in particular feels like a source of a lot of potential future churn, that may be better isolated in a separate project?
> 
> I'm open to discussion on the above topic.

@kmanning Maybe I'm a bit biased, but I'd say this should be useful to anyone who uses GitHub (either .com, cloud, or self-hosted GHE). My personal feeling is that a terraform-focused library that runs ``plan``s for a PR but doesn't comment on the PR, is about as useless to me as a unit test or code coverage tool that doesn't send its results back to the PR. A pull request should be the canonical place for information about a given change, and for a terraform project, the ``plan`` output is arguably the single most important piece of information about that change. I'll also say that, having interacted with other terraform users both in the general community and on other teams within my company, the ability to automatically run a ``plan`` for a PR and then expose the results as a comment on the PR seems to be one of the main "killer features" that gets people interested.

docs/TerraformPlanResultsPR.md
jleopold28: yes, changed it

Issue #102, opened 2019-09-03
Support target when running terraform plan and apply


kmanning:
* Support the ability to run apply on selective resource targets

kmanning:
Sample Code:

```
// Jenkinsfile
@Library(['terraform-pipeline']) _

Jenkinsfile.init(this, env)
TargetPlugin.init() // Optionally limit plan/apply to specific targets

def validate = new TerraformValidateStage()

def deployQA = new TerraformEnvironmentStage('qa')
def deployUat = new TerraformEnvironmentStage('uat')
def deployProd = new TerraformEnvironmentStage('prod')

// Pipeline can now be built with "Build with Parameters"
//     New 'target' option in parameter list can limit plan/apply to specific targets
//     By default, builds should skip the 'target' option, and build as-normal
validate.then(deployQa)
        .then(deployUat)
        .then(deployProd)
        .build()
```

PR #229, opened 2020-06-11
Support target for plan/apply

jleopold28:
Implementation for https://github.com/manheim/terraform-pipeline/issues/102

Supporting the `-target` parameter via Build Parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,6 @@ The example above gives you a bare-bones pipeline, and there may be Jenkinsfile 
 * [ConfirmApplyPlugin](./docs/ConfirmApplyPlugin.md): pause and review the plan, before applying any changes.
 * [ConditionalApplyPlugin](./docs/ConditionalApplyPlugin.md): only allow apply on master branch.
 * [DefaultEnvironmentPlugin](./docs/DefaultEnvironmentPlugin.md): automatically set `TF_VAR_environment` variable.
-* [TargetPlugin](./docs/TargetPlugin.md): set `-target` parameter for terraform plan and apply.
 ### Credentials and Configuration Management
 * [CredentialsPlugin](./docs/CredentialsPlugin.md): Inject Jenkins credentials into your stages.
 * [FileParametersPlugin](./docs/FileParametersPlugin.md): Use properties files to inject environment-specific variables.
@@ -117,6 +116,7 @@ The example above gives you a bare-bones pipeline, and there may be Jenkinsfile 
 * [TerraformDirectoryPlugin](./docs/TerraformDirectoryPlugin.md): Change the default directory containing your terraform code.
 * [TerraformLandscapePlugin](./docs/TerraformLandscapePlugin.md): Enable terraform-landscape plan output.
 * [TerraformPlanResultsPR](./docs/TerraformPlanResultsPR.md): Use this to post Terraform plan results in the comments of a PR.
+* [TargetPlugin](./docs/TargetPlugin.md): set `-target` parameter for terraform plan and apply.
 
 ## Write your own Plugin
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -97,6 +97,7 @@ The example above gives you a bare-bones pipeline, and there may be Jenkinsfile
 * [ConfirmApplyPlugin](./docs/ConfirmApplyPlugin.md): pause and review the plan, before applying any changes.
 * [ConditionalApplyPlugin](./docs/ConditionalApplyPlugin.md): only allow apply on master branch.
 * [DefaultEnvironmentPlugin](./docs/DefaultEnvironmentPlugin.md): automatically set `TF_VAR_environment` variable.
+* [TargetPlugin](./docs/TargetPlugin.md): set `-target` parameter for terraform plan and apply.
```
kmanning, line 100: For the "Default Plugins" category - I had intended it to mean, "Plugins that are enabled with you doing anything."   That's not the case for TargetPlugin is it?  I hadn't planned on having it enabled by default.

jleopold28: Oh Sorry, I didn't know exactly what you mean by "Default Plugins". I would say this is NOT a default plugin.

docs/TargetPlugin.md
```diff
@@ -0,0 +1,27 @@
...
+Enable this plugin to run plan/apply on selective resource targets
+
+```
+// Jenkinsfile
+@Library(['terraform-pipeline']) _
+
+Jenkinsfile.init(this, env)
+TargetPlugin.init() // Optionally limit plan/apply to specific targets
+
+
```
kmanning, line 12: Nit picky - are you intended extra whitespace?

If it for clarify, I'm good.  If it's a typo, I'd prefer it be cleaned up.

jleopold28: Yeah I just had the extra lines in there for visibility. Mainly to show there is the 'init' part of the code followed by the actual running of the pipeline. I really don't feel too strongly either way. I can clean it up if you prefer.

src/TargetPlugin.groovy
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public void apply(TerraformApplyCommand command) {
+        def targets = Jenkinsfile.instance.getEnv().RESOURCE_TARGETS
+        if ("${targets}" != '' && targets != null) {
+            for (item in "${targets}".split(',')) {
+                command.withArgument("-target ${item}")
+            }
+        }
```
kmanning, line 29: This is opinion - you don't have to change the above.  It feels a little wonky.

1.  If condition and logic operators support short-circuiting - once the condition in the left of `&&` evaluates to false, you no longer have to test the right side.  If you're worried about `null`, which you appear to be based on the use of string interpolation (`"${targets}"`) - you have a couple of options.
2.  You can use the elvis operator when you assign `targets` to guarantee that `targets` is a non-null value.  eg: `ef targets = Jenkinsfile.instance.getEnv().RESOURCE_TARGETS ?: ''` https://groovy-lang.org/operators.html#_elvis_operator
3. Once you have a non-null value, `split` doesn't care about empty strings (I bring up empty string, because you explicitly test for it).  Once you can guarantee a non-null value, you can leverage the behavior of `split` to handle the empty string.

Combining all of that, you could have something more concise:

```
        def targets = Jenkinsfile.instance.getEnv().RESOURCE_TARGETS ?: ''
        targets.split(',').each { item -> command.withArgument("-target ${item}") }
```

README.md
kmanning: Brainstorming, and asking your opinion here. 

I think this could go under "Other" OR it seems like we're building out a set of plugins that are kind of terraform-centric.  The TargetPlugin, the DirectoryPlugin, these are augmenting/configuring the behavior of terraform commands themselves. Maybe this warrants it's own category?

Either way, this is no big deal, but it's worth continuing to shape the organization of the project/plugins as a whole.

src/TargetPlugin.groovy
jleopold28: This is a good idea, I'm going to make this change and test it out

jleopold28: Tested out and this works perfectly!

README.md
jleopold28: I moved it under "Other" for now

kmanning:
Looks great!

src/TargetPlugin.groovy
kmanning: Looks great!

kmanning:
Looks great, thanks!

Issue #230, opened 2020-06-14
AgentNode Support with PR Plugin


bondezbond:
With the current PR plugin. AgentNode doesn't have support for Terraform Landscape since it uses hashicorp default image. To add that ability we need to be able to build dockerfiles in plugin and pass in build parameters. Thoughts on the update would be something like this

```
....
def terraformVersion = TerraformPlugin.checkVersion()
AgentNodePlugin.withAgentDockerImage("terraform-ruby:${terraformVersion}", true)
               .withAgentDockerBuildOptions("--build-arg TERRAFORM_VERSION=${terraformVersion}")
               .withAgentDockerImageOptions("--entrypoint=''")
               .init()
....
```

bondezbond:
`AgentNodePlugin.withAgentDockerImage` will take two parameters. The first will be the image name and second will be a `true/false` parameter to build from dockerfile or pull image

kmanning:
Related to #193

PR #231, opened 2020-06-14
Support building dockerfile & fix Terraform PR args

bondezbond:
Solves #230 #232 

## AgentNodePlugin
* Added ability to build using `Dockerfile`
* Added ability to pass build-args

## TerraformPlanResultsPR
* Use `withArgument` to pass terraform arguments

### DIFF
--- a/docs/AgentNodePlugin.md
+++ b/docs/AgentNodePlugin.md
@@ -2,13 +2,40 @@
 
 This plugin allows you to run the terraform stages in a docker container. This **DOES NOT WORK WITH AWSSUME** you should be using iam_block with aws provider with terraform.
 
+### Using pre-build docker image
 ```
 // Jenkinsfile
 @Library(['terraform-pipeline']) _
 
 Jenkinsfile.init(this)
 
-AgentNodePlugin.withAgentDockerImage('hashicorp/terraform:0.10.2')
+def terraformVersion = TerraformPlugin.checkVersion()
+AgentNodePlugin.withAgentDockerImage("hashicorp/terraform:${terraformVersion}")
+               .withAgentDockerImageOptions("--entrypoint=''")
+               .init()
+
+def validate = new TerraformValidateStage()
+
+def deployQA = new TerraformEnvironmentStage('qa')
+def deployUat = new TerraformEnvironmentStage('uat')
+def deployProd = new TerraformEnvironmentStage('prod')
+
+validate.then(deployQA)
+        .then(deployUat)
+        .then(deployProd)
+        .build()
+```
+
+### Using Dockerfile in repo
+```
+// Jenkinsfile
+@Library(['terraform-pipeline']) _
+
+Jenkinsfile.init(this)
+
+def terraformVersion = TerraformPlugin.checkVersion()
+AgentNodePlugin.withAgentDockerImage("man-vikings/terraform-ruby:${terraformVersion}", true)
+               .withAgentDockerBuildOptions("--build-arg TERRAFORM_VERSION=${terraformVersion}")
                .withAgentDockerImageOptions("--entrypoint=''")
                .init()
 
--- a/src/AgentNodePlugin.groovy
+++ b/src/AgentNodePlugin.groovy
@@ -3,6 +3,8 @@ import static TerraformEnvironmentStage.ALL
 
 public class AgentNodePlugin implements TerraformValidateStagePlugin, TerraformEnvironmentStagePlugin {
     private static String dockerImage
+    private static boolean withDockerfile
+    private static String dockerBuildOptions
     private static String dockerOptions
 
     AgentNodePlugin() { }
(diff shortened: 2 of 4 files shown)


bondezbond:
@jleopold28 yup tested on local repo

dkolb:
Can we perhaps make this another method and accept an optional path to the docker file?

```
AgentNodePlugin.withAgentDockerImage('local/custom-docker-image:latest')
    .withDockerfile(true)
    .withDockerfilePath('path/to/Dockerfile'')
```
This would be a bit more flexible to various repo layouts and stick with our more descriptive option naming instead of just a `true` sitting at the end of a method call.

docs/AgentNodePlugin.md
```diff
@@ -2,13 +2,40 @@
...
 
+### Using pre-build docker image
 ```
 // Jenkinsfile
 @Library(['terraform-pipeline']) _
 
 Jenkinsfile.init(this)
 
-AgentNodePlugin.withAgentDockerImage('hashicorp/terraform:0.10.2')
+def terraformVersion = TerraformPlugin.checkVersion()
```
kmanning, line 12: Is this pertinent to the AgentNodePlugin?

docs/AgentNodePlugin.md
```diff
@@ -2,13 +2,40 @@
...
+```
+
+### Using Dockerfile in repo
+```
+// Jenkinsfile
+@Library(['terraform-pipeline']) _
+
+Jenkinsfile.init(this)
+
+def terraformVersion = TerraformPlugin.checkVersion()
```
kmanning, line 36: Same as above - is terraformVersion actually pertinent to the AgentNodePlugin?

src/TerraformPlanResultsPR.groovy
```diff
@@ -44,10 +44,12 @@ class TerraformPlanResultsPR implements TerraformPlanCommandPlugin, TerraformEnv
     @Override
     public void apply(TerraformPlanCommand command) {
         if (landscape) {
```
kmanning, line 46: Building landscape logic into this plugin is continuing to feel "wrong".  This PR is already uncovering that there's a weakness in TerraformCommands, and it's starting to address the weakness (by pushing `-out=tfplan` into an argument, where it belongs).  I think we need to take this further, and maybe start spelling out a `.withOutputRedirection` or something similar, for `2>plan.err` to live.  

If the stderr redirection was treated differently than a normal "suffix" (and as you're illustrating, it is different), then we could push the landscape logic out of this plugin as well.

^-- this doesn't need to be done as part of this particular PR, but it should be done before the PR Plugin is released.

src/AgentNodePlugin.groovy
```diff
@@ -36,10 +44,14 @@ public class AgentNodePlugin implements TerraformValidateStagePlugin, TerraformE
 
     public Closure addAgent() {
         return { closure ->
-            if (dockerImage) {
+            if (this.dockerImage && this.withDockerfile == false) {
                 docker.image(this.dockerImage).inside(this.dockerOptions) {
                     closure()
                 }
+            } else if (this.dockerImage && this.withDockerfile == true) {
```
kmanning, line 51: It's minor, but `withDockerfile == true`, `withDockerfile == false` feels wonky.  

You could collapse this logic into:
```
if (dockerImage) {
    if (withDockerfile) {
        docker.build(...)
    } else {
        docker.image(...)
    }
} else {
    ....
}
```
Or at the very least, not compare against the literals:
```
if (this.dockerImage && !this.withDockerfile) {
    ...
} else if (this.dockerImage && this.withDockerfile) {
    ...
}
```

kmanning:
Can you update the ChangeLogs to reference your change?

src/AgentNodePlugin.groovy
kmanning: Ah, I think this is why it feels wonky to me - this is  in the Sun Coding Standard - https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/sun_checks.xml#L171

kmanning:
Related to #193

docs/AgentNodePlugin.md
bondezbond: that isn't pertinent. I can remove that feature but I wanted to show. how to use use it with the terraform version. Cause just having a terraform version file won't pull the right image.

docs/AgentNodePlugin.md
bondezbond: same response as above

src/AgentNodePlugin.groovy
bondezbond: I can change it either way. I tend to write literal checks so you know exactly what the code is looking for

bondezbond:
@dkolb great suggestion. I'll update that option. Maybe I'll just do
```
    .withDockerfilePath('path/to/Dockerfile'')
```
instead of two functions? Cause to me if your passing DockerFilePath inherently that would mean `withDockerFile = true`, does that sound good?

bondezbond:
thanks for the review @kmanning i'll update everything you mentioned

dkolb:
@bondezbond Yeah that sounds perfect.

kmanning:
This PR was replaced by https://github.com/manheim/terraform-pipeline/pull/241.  Closing this without merging.

PR #245, opened 2020-07-15
Issue 193: Simplify implementation

kmanning:
* More for Issue #193 
* Repo slug can be constructed from the scm URL
* The base URL can be constructed from the scm URL
* Remove landscape functionality from the plugin - there's a separate Landscape plugin


kmanning:
Maybe docs should indicate that SCM needs to be configured as a Github project. It doesn't work otherwise.

kmanning:
Thoughts on renaming this plugin to something like GithubPullRequestPlugin?

jleopold28:
@kmanning If you want to rename, I personally like "GithubPRPlanPlugin". Also I wanted to point out that if you are using this plugin with the landscape plugin, users would need to make sure the landscape plugin is defined before the PR plugin.

kmanning:
Regarding the ordering, that was also pointed out in another context.  v5.8 includes an issue that explicitly spells out Plugin Order significance, since it applies to most plugins - https://github.com/manheim/terraform-pipeline/issues/206.

Issue #88, opened 2019-09-03
Add the ability to destroy environments through the pipeline


kmanning:
**Request**
Ability to destroy resources via terraform-pipeline.

**Why?**
Deploying at a predetermined time can cause "release" problems. One example is listed below.

> Auto Scaling activity failed 73 seconds ago with error: We currently do not have sufficient t3.medium capacity in the Availability Zone you requested (us-east-1b).

**How?**
AWS EC2 Capacity Reservation is requested for each AZ, X hours before the predetermined release time. After the release is complete, the AWS EC2 Capacity Reservation is cancelled.

> aws_ec2_capacity_reservation
> https://www.terraform.io/docs/providers/aws/r/ec2_capacity_reservation.html

kmanning:
Hrm - can you clarify?  How exactly would you expect to use this in a project?

Are you expecting the TerraformDestroyStage to accept an argument of a *particular* resource?  

As it stands, terraform-pipeline has the concept of entire 'environments', regardless of how many or how few resources exist within that environment.

Moreover, my (possibly incorrect) assumption, would be that the UseCase for this new Stage, would simply be to retire an entire application, or environment - just once.  Something like:

```
def destroyQa = new TerraformDestroyStage('qa')
def destroyUat = new TerraformDestroyStage('uat')
def destroyProd = new TerraformDestroyStage('prod')

destroyQa.then(destroyUat).then(destroyProd)
```

I'd run this once, and then delete all the source code.

^--- What I'm describing doesn't sound like what you're hoping to get.

kmanning:
**terraform**
```
# AWS
provider "aws" {
  version = "~> 2.6"
  region  = "us-east-1"
}

locals {
  instance_count = 5
}

resource "aws_ec2_capacity_reservation" "us-east-1a" {
  instance_type     = "t3.medium"
  ebs_optimized     = true
  instance_platform = "Linux/UNIX"
  availability_zone = "us-east-1a"
  instance_count    = "${local.instance_count}"
}

resource "aws_ec2_capacity_reservation" "us-east-1b" {
  instance_type     = "t3.medium"
  ebs_optimized     = true
  instance_platform = "Linux/UNIX"
  availability_zone = "us-east-1b"
  instance_count    = "${local.instance_count}"
}

resource "aws_ec2_capacity_reservation" "us-east-1c" {
  instance_type     = "t3.medium"
  ebs_optimized     = true
  instance_platform = "Linux/UNIX"
[...]

Issue #24, opened 2019-09-03
ConfirmApplyPlugin - allow message customization


kmanning:
* Allow it to be enabled/disabled
* Allow the text of the prompts to be overridden

Example:
```
...
ConfirmApplyPlugin.disable()
...
```

The human confirmation step for every environment should now go straight from plan to apply, without prompting the user.

Do we care about skipping *some* environments and not others?

kmanning:
Acceptance Criteria:

| Given | When | Then |
| --- | --- | --- |
| I override the confirm message | my pipeline runs | I see the new message asking me to confirm before applying the terraform plan | 
| I override the Ok message | my pipeline runs | I see the new label on the OK button  to confirm |
| I override the submitter | my pipeline runs | (where is this even used...?) |
| I disable the ConfirmApplyPlugin | my pipeline runs | I am no longer prompted  to confirm - the apply runs automatically after the plan |

kmanning:
The disable functionality seems potentially dangerous.  Consider separating that into another issue.  If we move forward with it, it should get a clear warning about what will happen.

kmanning:
Considering adding the environment name to the confirmation button

kmanning:
Removing disable functionality, and pushing it to another Issue #264

PR #251, opened 2020-07-16
Adding support for Terraform Destroy. Adding ConfirmApply customization

jleopold28:
Support for TF destroy (https://github.com/manheim/terraform-pipeline/issues/88)

Also adding support to customize the ConfirmApply messages. (https://github.com/manheim/terraform-pipeline/issues/24)


src/ConditionalApplyPlugin.groovy
```diff
@@ -1,18 +1,22 @@
...
     }
 
     @Override
     public void apply(TerraformEnvironmentStage stage) {
         stage.decorateAround(CONFIRM, onlyOnExpectedBranch())
         stage.decorateAround(APPLY, onlyOnExpectedBranch())
+        if (stage instanceof TerraformDestroyStage){
+            stage.decorateAround(DESTROY, onlyOnExpectedBranch())
+        }
     }
```
kmanning, line 20: There's two ways you can avoid the `instanceof TerraformDestroyStage`.
1. If you define a `public void apply(TerraformDestroyStage stage)` method, Groovy will be smart enough to use that one for TerraformDestroyStages, and use this one for normal TerraformEnvironmentStages.  That is, you let the compiler implicitly do the `if instanceof TerraformDestroy` on your behalf.
2. We don't care whether or not you're a normal TerraformEnvironmentStage or a TerraformDestroyStage.  Either way, we do the same thing, and decorate the CONFIRM and APPLY.  This means that in a TerraformDestroyStage 'APPLY' is now the time that we run the `terraform destroy` command instead of the `terraform apply` command.

src/ConfirmApplyPlugin.groovy
```diff
@@ -13,19 +14,24 @@ class ConfirmApplyPlugin implements TerraformEnvironmentStagePlugin {
...
     public void apply(TerraformEnvironmentStage stage) {
+        command_name = "apply"
+        if (stage instanceof TerraformDestroyStage){
+            command_name = "destroy"
+        }
+
         if (enabled) {
-            stage.decorate(CONFIRM, addConfirmation())
+            stage.decorate(CONFIRM, addConfirmation(command_name))
         }
```
kmanning, line 24: Here again, we can avoid the `instanceof`, and in fact, it'll give us an added bonus.  We could leave `apply(TerraformEnvironmentStage stage)` unmodified, and define a new `apply(TerraformDestroyStage)`.  It could get a *FULLY* custom message "*DESTORY* *DESTROY* YOU ARE GOING TO DESTROY YOUR ENVIRONMENT *DESTROY* *DESTROY*" <-- I'm on the fence as to whether that's necessary, but I am calling it out as a possibility.

src/TargetPlugin.groovy
```diff
@@ -27,6 +28,15 @@ class TargetPlugin implements TerraformPlanCommandPlugin, TerraformApplyCommandP
...
+    @Override
+    public void apply(TerraformDestroyCommand command) {
+        def targets = Jenkinsfile.instance.getEnv().RESOURCE_TARGETS ?: ''
+        targets.split(',')
+               .collect { item -> item.trim() }
+               .findAll { item -> item != '' }
+               .each { item -> command.withArgument("-target ${item}") }
+    }
+
     @Override
```
kmanning, line 40: Does the TargetPlugin apply in the case of destroy? If you wanted to remove a single resource without touching anything else, couldn't you use a normal pipeline, remove your resource from your code, and target just the removed resource?

If we can avoid changing plugins at all, I would love that.

src/TerraformDestroyCommand.groovy
```diff
@@ -0,0 +1,88 @@
+class TerraformDestroyCommand {
+    private boolean input = false
```
kmanning, line 2: I can't say yet whether or not it might make sense to say `class TerraformDestoryCommand extends TerraformApplyCommand`?

src/TerraformDestroyStage.groovy
```diff
@@ -0,0 +1,52 @@
...
+                            decorations.apply(CONFIRM) {
+                                echo "Approved"
+                            }
+                        }
+                    }
+
+                    decorations.apply("Around-${DESTROY}") {
+                        stage("${DESTROY}-${environment}") {
+                            decorations.apply(DESTROY) {
+                                sh destroyCommand.toString()
```
kmanning, line 43: Here's where we could be "clever".  Instead of creating a new hook-point of "DESTROY", we *could* adapt/reuse "APPLY".  Then "everything just works" in terms of plugins, and perhaps no plugins would have to be modified to work with TerraformDestroy?

src/TerraformEnvironmentStage.groovy
```diff
@@ -1,14 +1,14 @@
...
+    protected StageDecorations decorations
+    protected TerraformInitCommand initCommand
+    protected TerraformPlanCommand planCommand
+    protected TerraformApplyCommand applyCommand
+    protected localPlugins
 
-    private static final DEFAULT_PLUGINS = [ new ConditionalApplyPlugin(), new ConfirmApplyPlugin(), new DefaultEnvironmentPlugin() ]
-    private static globalPlugins = DEFAULT_PLUGINS.clone()
+    protected static final DEFAULT_PLUGINS = [ new ConditionalApplyPlugin(), new ConfirmApplyPlugin(), new DefaultEnvironmentPlugin() ]
+    protected static globalPlugins = DEFAULT_PLUGINS.clone()
```
kmanning, line 12: Now that you've done that, you just highlighted the fact that many of these variables never needed to be instance variables to begin with.

Rather than make initCommand/planCommand/applyCommand protected, you could simply delete them, and make them local only to the pipelineConfiguration method.  In which case TerraformDestroy would have no idea about them (which is good) and would simply define it's own.

environment already has a public getEnvironment() method, it would not need to be made public.  jenkinsfile can be accessed globally, it too would not need to be made protected.

applyPlugins manipulates the plugins as needed, none of the plugin variables would need to be made protected.

kmanning: I feel like you've highlighted a code smell as well.  I think you're showing that maybe pipelineConfiguration should perhaps be implemented in some other way.  In fact, you might have just hit the nail on the head.  If pipelineConfiguration were defined by a class/object that could be swapped out, that might in fact solve *everything*.  We wouldn't need a new TerraformDestoryStage at all.

src/ConfirmApplyPlugin.groovy
```diff
@@ -14,18 +15,18 @@ class ConfirmApplyPlugin implements TerraformEnvironmentStagePlugin {
...
         return { closure ->
             // ask for human input
             try {
                 timeout(time: 15, unit: 'MINUTES') {
                     input(
-                        message: 'Are you absolutely sure the plan above is correct, and should be IMMEDIATELY DEPLOYED via "terraform apply"?',
-                        ok: 'Run terraform APPLY now',
+                        message: "Are you absolutely sure the plan above is correct, and should be IMMEDIATELY DEPLOYED via \"terraform ${command_name}\"?",
+                        ok: "Run terraform ${command_name.toUpperCase()} now",
                         submitterParameter: 'approver'
```
kmanning, line 30: Ooohh, this is super clever.  What if  we took it a  step further?

```
class ConfirmApplyPlugin...
   ...
   private  confirmMessage = 'Are you absolutely sure...' <--  the stuff that's currently hard coded.
   ...
        input (
            message: confirmMessage,
            ok: okMessage,
            submitterParameter: submitterParameter
        )
    ....
    public static withConfirmMessage(String newMessage) { this.confirmMessage = newMessage }
    public static withOkMessage(String newMessage) { this.okayMessage = newMessage }
    public static withSubmitterParameter(String newParam) { this.submitterParameter = newParam }
```

Now this has nothing to do with the DestroyPlugin.  *ANY* plugin that may need to manipulate confirmations can do so.

src/DestroyPlugin.groovy
```diff
@@ -0,0 +1,16 @@
...
+        ConfirmApplyPlugin.withCommand("destroy")
+        TerraformEnvironmentStage.addPlugin(plugin)
+    }
+
+    @Override
+    public void apply(TerraformEnvironmentStage stage) {
+        stage.withStrategy(new DestroyStrategy())
+    }
+
+}
```
kmanning, line 16: What you did above gave me a new idea!! What if...

```
class DestroyPlugin implements TerraformEnvironmentStagePlugin {

    public static void init() {
        DestroyPlugin plugin = new DestroyPlugin()

        // Need to set command name to "destroy"
        ConfirmApplyPlugin.withConfirmationMessage("YOU ARE ABOUT TO DESTORY YOUR ENVIORNMENT")
        TerraformPlanCommand.withDestroy()
        TerraformApplyCommand.withDestroy()
    }
}
```

```
class TerraformPlanCommand {
...
    public static withDestroy() {
        withArgument('-destory')
    }
}
```

```
class TerraformApplyCommand {
...
    public static withDestroy() {
         this.command = 'destroy'
    }
}
```

1. We would no longer need to modify TerraformEnvironmentStage *at all*.  
2. We wouldn't need TerraformDestroyStage.  
3. We wouldn't need a new TerraformDestoryCommand
4. *ALL* existing plugins would "just work"

kmanning: The above could even be implemented without adding new `withDestroys()` methods.  And in fact, not creating these new methods might be the better/more-conservative approach.

vincentclee:
This is some great work @jleopold28 ! Thoughts on adding `-refresh=false` to the default options for destroy?

**Common Scenario**
https://github.com/hashicorp/terraform/issues/15386#issuecomment-588295351

@kmanning

jleopold28:
@vincentclee thanks for testing! I don't think `-refresh=false` should be a default option, but I am in favor of allowing it to be configured.

docs/DestroyPlugin.md
```diff
@@ -0,0 +1,73 @@
...
+def destroyUat = new TerraformEnvironmentStage('uat')
+def destroyProd = new TerraformEnvironmentStage('prod')
+
+// Destroy the environments
+validate.then(destroyQa)
+        .then(destroyUat)
+        .then(destroyProd)
+        .build()
+```
+
```
kmanning, line 70: I'm not sure I'm a fan of this.  The premise of the library is that once you  call `.build()`, no other code should be executed.  I think it might be a necessary constraint in order to be to effectively create Declarative Pipelines (the syntax is really strict).  Note: [README](https://github.com/manheim/terraform-pipeline#how-to-use) `The design of this library is influenced by the Builder Pattern - your pipeline has been configured, but hasn't been constructed just yet. Finalize and create your pipeline by calling the build() method. This should only be done once - no code should come after calling this method.`


At any rate, I don't think I'm ready to concede on the `.build()` rule just yet.  

Can we  separate this PR into UseCases?  There's 1 UseCase of just terminating all the environments, which is your first example.  Can we do a separate PR to implement the other UseCase of deploying+destroying?  I'd like to talk more about how we  could implement that.

src/ConditionalApplyPlugin.groovy
```diff
@@ -13,6 +14,7 @@ public class ConditionalApplyPlugin implements TerraformEnvironmentStagePlugin {
     public void apply(TerraformEnvironmentStage stage) {
         stage.decorateAround(CONFIRM, onlyOnExpectedBranch())
         stage.decorateAround(APPLY, onlyOnExpectedBranch())
+        stage.decorateAround(DESTROY, onlyOnExpectedBranch())
     }
```
kmanning, line 18: If we only attack UseCase 1 in this PR (destroy-only), we no longer need these changes.

src/DefaultStrategy.groovy
```diff
@@ -0,0 +1,53 @@
...
+                                sh initCommand.toString()
+                                sh applyCommand.toString()
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
```
kmanning, line 53: Same as above: "If we only attack UseCase 1 in this PR (destroy-only), we no longer need these changes."

src/DestroyStrategy.groovy
```diff
@@ -0,0 +1,53 @@
...
+                            decorations.apply(DESTROY) {
+                                sh destroyCommand.toString()
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
```
kmanning, line 53: Same "If we only attack UseCase 1 in this PR (destroy-only), we no longer need these changes."

src/TerraformEnvironmentStage.groovy
```diff
@@ -46,46 +45,13 @@ class TerraformEnvironmentStage implements Stage {
...
-                                sh initCommand.toString()
-                                sh applyCommand.toString()
-                            }
-                        }
-                    }
-                }
-            }
-        }
+        return strategy.createPipelineClosure(environment, decorations)
     }
```
kmanning, line 55: Same "If we only attack UseCase 1 in this PR (destroy-only), we no longer need these changes."

kmanning:
@vincentclee @jleopold28 ?

In regards to the `-refresh=false`, does anything need to be done from a code  perspective?  Would `TerraformApply.withArgument('-refresh=false')` work?

kmanning:
Oh!  I also finally published the new Contributor instructions.  Let's do away with long-lived dev branches, and queue work in master instead.  It's simpler that way.  (Ie: you can make your PR against master, and I'll delete v5.9-dev - there's nothing in there at the moment).  https://github.com/manheim/terraform-pipeline#how-to-contribute

jleopold28:
@kmanning I updated the base to `master`. I also removed the section in the docs talking about "Destroy after Deploy". We can discuss that as a future piece of work.

jleopold28:
> @vincentclee @jleopold28 ?
> 
> In regards to the `-refresh=false`, does anything need to be done from a code perspective? Would `TerraformApply.withArgument('-refresh=false')` work?

That would work. However, I don't think that should be the default behavior...

kmanning:
You don't think "-refresh=false" should be the default behavior, or you don't think users should *need* to use `.withArgument` in order to get this to work?

jleopold28:
> You don't think "-refresh=false" should be the default behavior, or you don't think users should _need_ to use `.withArgument` in order to get this to work?

@kmanning I think users should have to use `withArgument` or something similar to add that option. 

When doing a destroy, I think we should always refresh the state before executing a destroy. If users are hitting a bug that requires them to add `refresh=false`, they should specify that with arguments during the plugin init.

src/TerraformEnvironmentStage.groovy
```diff
@@ -14,6 +12,7 @@ class TerraformEnvironmentStage implements Stage {
     public static final String PLAN = 'plan'
     public static final String CONFIRM = 'confirm'
     public static final String APPLY = 'apply'
+    public static final String DESTROY = 'destroy'
```
kmanning, line 15: This new state of "DESTROY".  How will this behavior with the existing plugins that are part of  the library (or worse, plugins that are in completely external libraries?)

eg: 

* https://github.com/manheim/terraform-pipeline/blob/74de97b89fdedf087fde028686b1e6530a30aed2/src/ParameterStoreExecPlugin.groovy#L14-L20
* https://github.com/manheim/terraform-pipeline/blob/74de97b89fdedf087fde028686b1e6530a30aed2/src/AnsiColorPlugin.groovy#L11-L14
* https://github.com/manheim/terraform-pipeline/blob/817be51f85223d535120fb7d17e35698f34d8f17/src/ParameterStoreBuildWrapperPlugin.groovy#L17-L29
* https://github.com/manheim/terraform-pipeline/blob/00e3eed6cfcf8e4ad40a68a323f116202cda90da/src/CrqPlugin.groovy#L11-L16

docs/DestroyPlugin.md
```diff
@@ -0,0 +1,50 @@
...
+When using this plugin, your pipeline will look something like this:
+
+![DestroyPluginPipeline](../images/destroy-pipeline.png)
+
```
kmanning, line 35: Nice touch on the screenshot - thank you for that.

src/TerraformEnvironmentStage.groovy
jleopold28: @kmanning I am making some changes now. I will have to add `stage.decorate(DESTROY, someMethod(options))` to the above plugins. This will add functionality if DESTROY is enabled. Otherwise this will not change the existing functionality.

kmanning:
Thanks for accounting for the other plugins.  I'm still concerned about the maintainability of the existing PLAN/APPLY hooks, and I'm concerned about introducing more problems with the new DESTROY hook.  But I think that can be tackled in a separate PR/refactor.

kmanning:
:shipit:

Issue #162, opened 2019-09-23
Support "plan only" on master


kmanning:
* Allow a user to run the master branch, but as 'plan only'
* When enabled, the confirm and apply steps will be skipped on the master branch
* Allow the  user to indicate  that the pipeline should fail if un-applied changes are found in any environment.
* UseCase:
    * Each week, I want to trigger all my pipelines to look for un-applied changes
    * When triggered, 'plan' should be run across each environment in the pipeline
    * If any environment has un-applied changes, the pipeline should fail-fast
* Dev Notes: Generally speaking, this should be implemented as a Plugin.  Build this behavior into https://github.com/manheim/terraform-pipeline/blob/master/docs/ConditionalApplyPlugin.md if it makes sense for the behavior to live here.  If it doesn't make sense, build a new Plugin.

kmanning:
This affects branches other than master.  Should that be the case?

PR #259, opened 2020-07-28
Issue #162: Support "Plan only"

jleopold28:
* Support for https://github.com/manheim/terraform-pipeline/issues/162

* Added support for adding parameters to Jenkinsfile.
* Modified TargetPlugin to use new parameter methods
* Updated tests for TargetPlugin


src/DestroyStrategy.groovy
```diff
@@ -33,6 +33,7 @@ class DestroyStrategy {
             node(jenkinsfile.getNodeName()) {
                 deleteDir()
                 checkout(scm)
+                properties([parameters(params)])
```
kmanning, line 37: This is making me really uncomfortable from a maintainability perspective.  I don't want to have to update every strategy from now and into the future to make stuff like this work.

Can we approach this differently?  This could be a decoration around ALL, couldn't it?  

^---These statements apply to the DefaultStrategy as well.

src/PlanOnlyStrategy.groovy
```diff
@@ -0,0 +1,37 @@
+import static TerraformEnvironmentStage.ALL
+import static TerraformEnvironmentStage.PLAN
+
+class PlanOnlyStrategy {
+
```
kmanning, line 5: This is making me really uncomfortable from a maintainability perspective.  I've already been chewing on removing the Strategy implementation from the 5.9 release for this reason.  I think more work needs to go into the Strategy idea before we start building plugins on it.

src/PlanOnlyStrategy.groovy
```diff
@@ -0,0 +1,37 @@
...
+                    stage("${PLAN}-${environment}") {
+                        decorations.apply(PLAN) {
+                            sh initCommand.toString()
+                            def status = sh(returnStatus: true, script: "set -o pipefail;" + planCommand.toString())
+                            if ( status != 0 ) {
+                                echo "ERROR: Expected NO CHANGES to terraform resources."
+                                currentBuild.result = 'FAILURE'
+                            }
+                        }
+                    }
```
kmanning, line 32: Can this be implemented as a decoration around Plan?

We can also skip the apply in the same way that the ConditionalApplyPlugin does.  These approaches don't require us to alter the internal  structure of TerraformEnvironmentStage.

src/TargetPlugin.groovy
```diff
@@ -28,22 +32,5 @@ class TargetPlugin implements TerraformPlanCommandPlugin, TerraformApplyCommandP
...
-                string(name: 'RESOURCE_TARGETS', defaultValue: '', description: 'comma-separated list of resource addresses to pass to plan and apply "-target=" parameters'),
-            ]
-            def props = [
-                parameters(params)
-            ]
-            properties(props)
-
-            closure()
-        }
-    }
```
kmanning, line 47: I think I actually prefer this implementation. It doesn't require the Plugin to know about the internal structure of the TerraformEnvironmentStage.

src/DestroyStrategy.groovy
kmanning: Hrm, this is the issue with multiple plugins trying to set properties isn't it?

If multiple plugins want to inject properties, they overwrite each other?

jleopold28: I'm not sure what you mean by "I don't want to have to update every strategy from now and into the future to make stuff like this work". You wouldn't need to update every strategy going forward.

When you write a new strategy, you would just copy the code from defaultStrategy (which already has the `properties([parameters(params)])` line. 

I had experimented with trying a decoration around ALL. However, I was hitting a race condition where the parameters would not be defined when running the pipeline (since they were appended as parameters after the building of the pipeline)

src/PlanOnlyStrategy.groovy
jleopold28: I'll have to play around to see if this can be a decoration around Plan. The problem with that is I need to capture the exit code from the plan command.

src/TargetPlugin.groovy
jleopold28: With this implementation, I cannot append to the parameters. So any plugins that add parameters will overwrite eachother.

src/DestroyStrategy.groovy
kmanning: Has this been tested against Declarative pipelines?

jleopold28: > Hrm, this is the issue with multiple plugins trying to set properties isn't it?
> 
> If multiple plugins want to inject properties, they overwrite each other?

yes they overwrite. Thats why I wanted to define a list of params which we can append to.

jleopold28: > Has this been tested against Declarative pipelines?

No not yet

src/Jenkinsfile.groovy
```diff
@@ -74,6 +75,7 @@ class Jenkinsfile {
     }
 
     public static build(Closure closure) {
+        original.ApplyJenkinsfileClosure(createParamClosure())
```
jleopold28, line 78: I'm getting a little lost on how to support adding in the parameters closure. I think the way I have it set up now, it is adding the params before each stage, which is not what we want. 

If we want to add the parameters ONLY once (before the stages are even defined), I was thinking about setting up something like this

```
    public static void build(List<Stage> stages) {
        if (!declarative) {
            original.ApplyJenkinsfileClosure(createParamClosure())
            stages.each { Stage stage -> stage.build() }
        } else {
            if (pipelineTemplate == null) {
                this.pipelineTemplate = getPipelineTemplate(stages)
            }
            pipelineTemplate.call(createParamClosure())
            pipelineTemplate.call(stages)
        }
    }
```

However this is causing the BuildGraphTest to fail with `ava.lang.NullPointerException: Cannot invoke method ApplyJenkinsfileClosure() on null object`

Issue #254, opened 2020-07-24
Auto-convert git/ssh-urls to https with GithubPRPlanPlugin


jleopold28:
When attempting to clone an https repository (`https://github.com/blah/foo`), the pipeline actually pulls via the git protocol (`git://github.com/blah/foo`).

This causes a problem when we execute the curl command to post comments. 
See:
```
+ curl -H 'Authorization: token ****' -X POST -d @/home/jenkins/workspace/_terraform-pipeline-example_PR-4@tmp/body.txt -H 'Content-Type: application/json' -D comment.headers git://github.com/api/v3/repos/foo/terraform-pipeline-example/issues/4/comments
curl: (1) Protocol git not supported or disabled in libcurl
```

kmanning:
It looks like you're using the GithubOrg jenkins plugin, and you've added behavior to "Checkout over SSH".  As a result, the git url that the plugin is finding has the `git://` protocol instead of an `https://` protocol.

kmanning:
I think there are 2 possible options here:
1. We can advertise the `withRepoHost()` method, and you can override the github URL that is being used to check the project out.
2. We can auto-magically check and replace the `git://` protocol with `https://`.  <-- I think this is opening up a can of worms.  Since the user chose to check the repo out with SSH, we don't know if their github instance is available over http or https.

kmanning:
Have we validated that the PR link is constructed correctly for the public github?

kmanning:
Bumping this to v5.10, so we can cut a v5.9 release

jleopold28:
> Have we validated that the PR link is constructed correctly for the public github?

You mean posting the plan results to public github? I remember testing that when we did the initial release

PR #282, opened 2020-08-04
Issue #254: GithubPlanPR bug fix

jleopold28:
See https://github.com/manheim/terraform-pipeline/issues/254

### DIFF
--- a/docs/GithubPRPlanPlugin.md
+++ b/docs/GithubPRPlanPlugin.md
@@ -26,3 +26,18 @@ validate.then(deployQa)
         .then(deployProd)
         .build()
 ```
+
+## Changing Repo Host
+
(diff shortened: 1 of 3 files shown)


kmanning:
Can you add 2 unit tests for this?  

There's already a unit test that confirms that we're using the correct protocol.  Seems like we need an additional test case that says that if the protocol is git, it overrides it to https. https://github.com/manheim/terraform-pipeline/blob/master/test/GithubPRPlanPlugin.groovy#L165-L175

Likewise, let's add a unit test that shows that the default hostname is used, unless it is explicitly overridden.  https://github.com/manheim/terraform-pipeline/blob/master/test/GithubPRPlanPlugin.groovy#L164-L176

jleopold28:
@kmanning Unit test added

Issue #175, opened 2019-12-17
Pass terraform plan output to apply


bondezbond:
Current issue I know we've talked about is when running terraform plan output. The output file has relative paths to where it was ran. So if you ran apply on another node in jenkins and the paths where different. Things would break. 

I'm wondering what happens when you set the environment variable `TF_IN_AUTOMATION`? Does it remove all those relative path issues?

bondezbond:
Was reading this today https://learn.hashicorp.com/terraform/development/running-terraform-in-automation

bondezbond:
NVM. looking more through this. found this statement `Before running apply, obtain the archive created in the previous step and extract it at the same absolute path. `

I wonder if there's a better way to solve this?

jantman:
I _thought_ there was an open feature request on terraform to fix this, but apparently there isn't.

Extracting at the same absolute path is problematic in Jenkins, because (1) if you have multiple build nodes, you'd need to ensure that JENKINS_HOME is the same on all of them, and (2) if you're using GitHub Branch Source / Multibranch, I'm pretty sure the workspace paths are different for every build.

The only sure-fire way that I can think of to fix this is to run terraform inside a Docker container (i.e. with ``docker.image('image/name:tag').inside() { ... }``, and mount the workspace in to some reproducible path like ``/app``. But that's pretty ugly for multiple reasons, if you're not already doing it.

PR #294, opened 2020-09-01
Issue 175: Pass terraform plan file to apply

jleopold28:
Adding support for Issue #175

### DIFF
--- a/src/PassPlanFilePlugin.groovy
+++ b/src/PassPlanFilePlugin.groovy
@@ -26,7 +26,7 @@ class PassPlanFilePlugin implements TerraformPlanCommandPlugin, TerraformApplyCo
     @Override
     public void apply(TerraformApplyCommand command) {
         String env = command.getEnvironment()
-        command.withArgument("tfplan-" + env)
+        command.withDirectory("tfplan-" + env)
     }
 
     public Closure archivePlanFile(String env) {
--- a/src/PassPlanFilePlugin.groovy
+++ b/src/PassPlanFilePlugin.groovy
@@ -16,7 +16,8 @@ class PassPlanFilePlugin implements TerraformPlanCommandPlugin, TerraformApplyCo
 
     @Override
     public void apply(TerraformEnvironmentStage stage) {
-        stage.decorate(PLAN, savePlanFile(stage.getEnvironment()))
+        stage.decorate(PLAN,  archivePlanFile(stage.getEnvironment()))
+        stage.decorate(APPLY, downloadArchive(stage.getEnvironment()))
     }
 
     @Override
@@ -27,16 +28,32 @@ class PassPlanFilePlugin implements TerraformPlanCommandPlugin, TerraformApplyCo
 
     @Override
     public void apply(TerraformApplyCommand command) {
-        command.withArgument(planAbsolutePath)
+        String env = command.getEnvironment()
+        command.withArgument("tfplan-" + env)
     }
 
-    public Closure savePlanFile(String env) {
+    public Closure archivePlanFile(String env) {
         return { closure ->
             closure()
             String workingDir = pwd()
             String planFileName = workingDir + "/tfplan-" + env
             archiveArtifacts artifacts: "tfplan-" + env
             setAbsolutePath(planFileName)
+        }
+    }
+
+    public Closure downloadArchive(String env) {
+        return { closure ->
+            String jenkinsUrl  = Jenkinsfile.instance.getEnv()['JENKINS_URL']
+            String jobBaseName = Jenkinsfile.instance.getEnv()['JOB_BASE_NAME']
+            String branch      = Jenkinsfile.instance.getEnv()['BRANCH_NAME']
+            String buildNumber = Jenkinsfile.instance.getEnv()['BUILD_NUMBER']
+            String url = jenkinsUrl + "/job/" + jobBaseName + "/" + branch + "/" + buildNumber + "/artifact/tfplan-" + env
+            echo url
+            sh "wget ${url}"
+
+            closure()
+
         }
     }
 
--- a/src/PassPlanFilePlugin.groovy
+++ b/src/PassPlanFilePlugin.groovy
@@ -46,10 +46,17 @@ class PassPlanFilePlugin implements TerraformPlanCommandPlugin, TerraformApplyCo
     public Closure downloadArchive(String env) {
         return { closure ->
             String jenkinsUrl  = Jenkinsfile.instance.getEnv()['JENKINS_URL']
+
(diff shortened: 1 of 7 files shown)


src/PassPlanFilePlugin.groovy
```diff
@@ -0,0 +1,71 @@
...
+    public void apply(TerraformApplyCommand command) {
+        String env = command.getEnvironment()
+        command.withDirectory("tfplan-" + env)
+    }
+
+    public Closure archivePlanFile(String env) {
+        return { closure ->
+            closure()
+            echo "Archiving tfplan-${env} file"
+            archiveArtifacts artifacts: "tfplan-" + env
```
kmanning, line 36: Is there a reason to use `archiveArtifacts` vs `stash`?

src/PassPlanFilePlugin.groovy
```diff
@@ -0,0 +1,71 @@
...
+    public Closure downloadArchive(String env) {
+        return { closure ->
+            String jenkinsUrl  = Jenkinsfile.instance.getEnv()['JENKINS_URL']
+            String jobName     = Jenkinsfile.instance.getEnv()['JOB_NAME']
+            String buildNumber = Jenkinsfile.instance.getEnv()['BUILD_NUMBER']
+            String url = getArtifactUrl(jenkinsUrl, jobName, buildNumber, env)
+            echo "Downloading tfplan-${env} file from ${url}"
+
+            sh "wget -O tfplan-${env} ${url}"
+
```
kmanning, line 49: Same as above - is there a reason to use archive vs stash?  The above is  creating a dependency on `wget`.

kmanning:
The two  new closures aren't getting exercised in any test.  https://codecov.io/gh/manheim/terraform-pipeline/pull/294/diff?src=pr&el=tree#diff-c3JjL1Bhc3NQbGFuRmlsZVBsdWdpbi5ncm9vdnk=

Just for the sake of ensuring there  are no syntax errors, can you add 2 tests that do something as simple as exercising the closure, without validating anything.  The test name can be something as simple as "runsWithoutError".  Even without behavior assertions, merely executing the method under a unit test context will provide some benefit.

Eg: https://github.com/manheim/terraform-pipeline/blob/master/test/BuildStageTest.groovy#L18-L28

src/PassPlanFilePlugin.groovy
jleopold28: Nope, I changed it to stash

jleopold28:
> https://codecov.io/gh/manheim/terraform-pipeline/pull/294/diff?src=pr&el=tree#diff-c3JjL1Bhc3NQbGFuRmlsZVBsdWdpbi5ncm9vdnk=

@kmanning I added those tests. Let me know if there is anything else you want changed

src/PassPlanFilePlugin.groovy
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    public Closure unstashPlan(String env) {
+        return { closure ->
+            String planFile = "tfplan-" + env
+            echo "Unstashing ${planFile} file"
+            unstash planFile
+            closure()
+        }
```
kmanning, line 47: Nice.  The change to use stash/unstash looks great =).

test/PassPlanFilePluginTest.groovy
```diff
@@ -0,0 +1,125 @@
...
+            def expectedClosure = { -> }
+            def plugin = spy(new PassPlanFilePlugin())
+            doReturn(expectedClosure).when(plugin).unstashPlan()
+            def stage = mock(TerraformEnvironmentStage.class)
+
+            plugin.apply(stage)
+
+            verify(stage).decorate(anyString(), eq(expectedClosure))
+        }
+
```
kmanning, line 102: I see the new `runsStashPlan` runsUnstashPlan` tests - thanks for adding that!

I'm looking at  the code coverage however, and it doesn't seem to have changed. Digging into the details of the test, it looks like you're mocking the closures, and the closures are the lines of code missing coverage.

The closure  is  the thing that I was interested in getting coverage on.  If there were, say, a syntactic error, simply exercising the closure would catch that.  Instead of exercising `stashPlan` and `unstashPlan`  through `apply`, could we break it down to a more fine-grained unit test, exercise those methods directly?

Something along the lines of: 
```
class UnstashPlan {
        @Test
        void justExercisingNotValidatingBehavior() {
            def plugin = new PassPlanFilePlugin()

            def unstashClosure = plugin.unstashPlan('myenv')
            unstashClosure.call { -> }
        }
}
```

^-- if you did the above, we should see code coverage go up, because we're actually exercising the closure.

If we wanted to take this further, we could even validate *some* amount of behavior.  It's really important that the inner closure is executed - if calling the inner closure were mistakenly left off, it would have really detrimental impact on the behavior of the pipeline.  So we could build more safety into our unit tests with something like:

```
class UnstashPlan {
        @Test
        void executesPassedClosure() {
            def wasCalled = false
            def innerClosure = { -> wasCalled = true }
            def plugin = new PassPlanFilePlugin()

            def unstashClosure = plugin.unstashPlan('myenv')
            unstashClosure.call(innerClosure)

            assertTrue(wasCalled)
        }
```

jleopold28: @kmanning I tried implementing that second block but I'm getting some errors.

Any ideas what to do about `groovy.lang.MissingMethodException: No signature of method: PassPlanFilePlugin.echo() is applicable for argument types: (org.codehaus.groovy.runtime.GStringImpl) values: [Stashing tfplan-dev file]`

kmanning: That makes sense.  Your closure is running Jenkinsfile DSL methods, like echo.  

```
    public Closure unstashPlan(String env) {
        return { closure ->
            String planFile = "tfplan-" + env
            echo "Unstashing ${planFile} file"
            unstash planFile
            closure()
        }
```

`echo` isn't defined in  the plugin, and it's not defined in your test.  When executing a closure, when the VM comes across a method, it does a search.  See [here](https://groovy-lang.org/closures.html#_delegation_strategy):
1. `this`
2. `owner`
3. `delegate`

If the method can't be  found after traversing those objects, it throws the MissingMethodException you're seeing. 

In a "normal" terraform-pipeline, the owner or delegate would be the actual Jenkinsfile executing on Jenkins.  But you're just in a unit test.  So what you can do, is explicitly set a delegate, and use the DummyJenkinsfile to stub in for what would normally happen in Jenkins.

As an example, you can check  out this test: https://github.com/manheim/terraform-pipeline/blob/9bac14b75ce353fbbe7592e45a9ec786e5824ab9/test/TerraformEnvironmentStageTest.groovy#L168-L178

PR #295, opened 2020-09-01
added docs on withEnv & withGlobalEnv

### LINKED ISSUE #293: add documentation for `withGlobalEnv`
Document `withGlobalEnv` so people understand how to use it

bondezbond:
fixes #293

### DIFF
--- a/docs/TerraformEnvironmentStage.md
+++ b/docs/TerraformEnvironmentStage.md
@@ -1,4 +1,4 @@
-## [TerraformEnvironmentStagePlugin](../src/TerraformEnvironmentStagePlugin.groovy)
+## [TerraformEnvironmentStage](../src/TerraformEnvironmentStage.groovy)
 
 ### Pass in Global Environment variable
 
--- a/docs/TerraformEnvironmentStage.md
+++ b/docs/TerraformEnvironmentStage.md
@@ -1,0 +1,44 @@
+## [TerraformEnvironmentStagePlugin](../src/TerraformEnvironmentStagePlugin.groovy)
+
+### Pass in Global Environment variable
+
+You can pass in global environment variables too all stages in your pipeline using `withGlobalEnv`
+
+```
+// Jenkinsfile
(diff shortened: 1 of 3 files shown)


kmanning:
Can you update the CHANGELOG to include your change?

kmanning:
Otherwise, looks great, thanks!

docs/TerraformEnvironmentStage.md
```diff
@@ -0,0 +1,44 @@
+## [TerraformEnvironmentStagePlugin](../src/TerraformEnvironmentStagePlugin.groovy)
```
kmanning, line 1: Should this be `TerraformEnvironmentStage` and not `TerraformEnvironmentStagePlugin`?

bondezbond: so I struggled with that. I had just Stage before but to follow the pattern other plugins I used the plugin of it since the plugin calls EnvironmentStage

kmanning: Yeah, that's part of the reason why it's undocumented.  It's "odd" behavior.

But that convention aside, the documentation is explaining the `.withEnv`, and `.withGlobalEnv`, and neither of those methods exist on the code that's being linked to by the documentation.

kmanning: The README has a section [Additional Pipeline Stages](https://github.com/manheim/terraform-pipeline#additional-pipeline-stages).  What if instead of calling that "Additional", what if it were simply "Pipeline Stages".  At which point, we can add `TerraformEnvironmentStage` to the list, and link to your doc.  

Would that make for a better convention?

kmanning: The above would also solve the problem of making this documentation "find-able".

bondezbond: that works. I'll update the PR

bondezbond: updated

PR #302, opened 2020-09-16
AWSParameterStore - Support Global Parameters

bondezbond:
* add support for global parameters

### DIFF
--- a/src/ParameterStoreBuildWrapperPlugin.groovy
+++ b/src/ParameterStoreBuildWrapperPlugin.groovy
@@ -25,8 +25,7 @@ class ParameterStoreBuildWrapperPlugin implements TerraformValidateStagePlugin, 
     @Override
     public void apply(TerraformValidateStage stage) {
         globalParameterOptions.each { gp ->
-            stage.decorate(PLAN, addParameterStoreBuildWrapper(gp))
-            stage.decorate(APPLY, addParameterStoreBuildWrapper(gp))
+            stage.decorate(ALL, addParameterStoreBuildWrapper(gp))
         }
     }
 
--- a/src/ParameterStoreBuildWrapperPlugin.groovy
+++ b/src/ParameterStoreBuildWrapperPlugin.groovy
@@ -3,6 +3,7 @@ import static TerraformEnvironmentStage.APPLY
 
 class ParameterStoreBuildWrapperPlugin implements TerraformEnvironmentStagePlugin {
     private static globalPathPattern
+    private static ArrayList globalParameterOptions = []
     private static defaultPathPattern = { options -> "/${options['organization']}/${options['repoName']}/${options['environment']}/" }
 
     public static void init() {
@@ -11,6 +12,11 @@ class ParameterStoreBuildWrapperPlugin implements TerraformEnvironmentStagePlugi
 
     public static withPathPattern(Closure newPathPattern) {
         globalPathPattern = newPathPattern
+        return this
+    }
+
+    public static withGlobalParameter(String path, options=[]) {
+        globalParameterOptions << [path: path] + options
         return this
     }
 
@@ -26,6 +32,11 @@ class ParameterStoreBuildWrapperPlugin implements TerraformEnvironmentStagePlugi
 
         stage.decorate(PLAN, addParameterStoreBuildWrapper(options))
         stage.decorate(APPLY, addParameterStoreBuildWrapper(options))
+
+        globalParameterOptions.each { gp ->
+            stage.decorate(PLAN, addParameterStoreBuildWrapper(gp))
+            stage.decorate(APPLY, addParameterStoreBuildWrapper(gp))
+        }
     }
 
     String pathForEnvironment(String environment) {
@@ -48,6 +59,7 @@ class ParameterStoreBuildWrapperPlugin implements TerraformEnvironmentStagePlugi
         def parameterStoreOptions = defaultOptions + options
 
         return { closure ->
+            // sh "echo ${parameterStoreOptions}" //DEBUG
             withAWSParameterStore(parameterStoreOptions) {
                 closure()
             }
@@ -56,5 +68,6 @@ class ParameterStoreBuildWrapperPlugin implements TerraformEnvironmentStagePlugi
 
     public static reset() {
         globalPathPattern = null
+        globalParameterOptions = []
     }
 }
--- a/src/ParameterStoreBuildWrapperPlugin.groovy
+++ b/src/ParameterStoreBuildWrapperPlugin.groovy
@@ -68,7 +68,6 @@ class ParameterStoreBuildWrapperPlugin implements TerraformValidateStagePlugin, 
         def parameterStoreOptions = defaultOptions + options
 
         return { closure ->
-            // sh "echo ${parameterStoreOptions}" //DEBUG
             withAWSParameterStore(parameterStoreOptions) {
                 closure()
             }
--- a/src/ParameterStoreBuildWrapperPlugin.groovy
+++ b/src/ParameterStoreBuildWrapperPlugin.groovy
@@ -4,7 +4,7 @@ import static TerraformEnvironmentStage.APPLY
 
 class ParameterStoreBuildWrapperPlugin implements TerraformValidateStagePlugin, TerraformEnvironmentStagePlugin {
     private static globalPathPattern
-    private static ArrayList<Map> globalParameterOptions = []
+    private static List globalParameterOptions = []
     private static defaultPathPattern = { options -> "/${options['organization']}/${options['repoName']}/${options['environment']}/" }
 
     public static void init() {
@@ -31,16 +31,25 @@ class ParameterStoreBuildWrapperPlugin implements TerraformValidateStagePlugin, 
 
     @Override
     public void apply(TerraformEnvironmentStage stage) {
-        def environment                    = stage.getEnvironment()
-        ArrayList<Map> allParameterOptions = []
+        def environment = stage.getEnvironment()
+        List options    = getParameterOptions(environment)
 
-        allParameterOptions << getEnvironmentParameterOptions(environment)
-        allParameterOptions.addAll(globalParameterOptions)
-
-        allParameterOptions.each { apo ->
-            stage.decorate(PLAN, addParameterStoreBuildWrapper(apo))
-            stage.decorate(APPLY, addParameterStoreBuildWrapper(apo))
+        options.each { option ->
+            stage.decorate(PLAN, addParameterStoreBuildWrapper(option))
+            stage.decorate(APPLY, addParameterStoreBuildWrapper(option))
         }
+    }
+
+    List getParameterOptions(String environment) {
+        List options = []
+        options.add(getEnvironmentParameterOptions(environment))
+        options.addAll(getGlobalParameterOptions())
+
+        return options
+    }
+
+    List getGlobalParameterOptions() {
+        return this.globalParameterOptions
     }
 
     Map getEnvironmentParameterOptions(String environment) {
@@ -52,10 +61,10 @@ class ParameterStoreBuildWrapperPlugin implements TerraformValidateStagePlugin, 
 
     String pathForEnvironment(String environment) {
         String organization = Jenkinsfile.instance.getOrganization()
-        String repoName = Jenkinsfile.instance.getRepoName()
-        def patternOptions = [ environment: environment,
-                               repoName: repoName,
-                               organization: organization ]
+        String repoName     = Jenkinsfile.instance.getRepoName()
+        def patternOptions  = [ environment: environment,
+                                repoName: repoName,
+                                organization: organization ]
 
         def pathPattern = globalPathPattern ?: defaultPathPattern
 
--- a/src/ParameterStoreBuildWrapperPlugin.groovy
+++ b/src/ParameterStoreBuildWrapperPlugin.groovy
@@ -2,7 +2,7 @@ import static TerraformValidateStage.ALL
 import static TerraformEnvironmentStage.PLAN
 import static TerraformEnvironmentStage.APPLY
 
-class ParameterStoreBuildWrapperPlugin implements TerraformEnvironmentStagePlugin {
+class ParameterStoreBuildWrapperPlugin implements TerraformValidateStagePlugin, TerraformEnvironmentStagePlugin {
     private static globalPathPattern
     private static ArrayList globalParameterOptions = []
     private static defaultPathPattern = { options -> "/${options['organization']}/${options['repoName']}/${options['environment']}/" }
--- a/src/ParameterStoreBuildWrapperPlugin.groovy
+++ b/src/ParameterStoreBuildWrapperPlugin.groovy
@@ -49,7 +49,7 @@ class ParameterStoreBuildWrapperPlugin implements TerraformValidateStagePlugin, 
     }
 
     List getGlobalParameterOptions() {
-        return this.globalParameterOptions
+        return globalParameterOptions
     }
 
     Map getEnvironmentParameterOptions(String environment) {
--- a/src/ParameterStoreBuildWrapperPlugin.groovy
+++ b/src/ParameterStoreBuildWrapperPlugin.groovy
@@ -4,7 +4,7 @@ import static TerraformEnvironmentStage.APPLY
 
 class ParameterStoreBuildWrapperPlugin implements TerraformValidateStagePlugin, TerraformEnvironmentStagePlugin {
     private static globalPathPattern
-    private static ArrayList globalParameterOptions = []
+    private static ArrayList<Map> globalParameterOptions = []
     private static defaultPathPattern = { options -> "/${options['organization']}/${options['repoName']}/${options['environment']}/" }
 
     public static void init() {
@@ -17,7 +17,7 @@ class ParameterStoreBuildWrapperPlugin implements TerraformValidateStagePlugin, 
         return this
     }
 
-    public static withGlobalParameter(String path, options=[]) {
+    public static withGlobalParameter(String path, Map options = [:]) {
         globalParameterOptions << [path: path] + options
         return this
     }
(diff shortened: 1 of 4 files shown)


bondezbond:
still need to add docs & changelog

src/ParameterStoreBuildWrapperPlugin.groovy
```diff
@@ -48,6 +59,7 @@ class ParameterStoreBuildWrapperPlugin implements TerraformEnvironmentStagePlugi
         def parameterStoreOptions = defaultOptions + options
 
         return { closure ->
+            // sh "echo ${parameterStoreOptions}" //DEBUG
```
kmanning, line 62: Let's remove this, and make an issue to generally support informational and debugging messages.  This isn't the only change that would benefit from the  ability to toggle "logging" on and off.

kmanning:
Looks like code coverage  went down.  You modified the `apply` method, but there are no unit tests exercising that.  Can you add a unit test that exercises the new code that you added?  The new unit tests should run apply against a Stage, and should have global parameter paths, so that we exercise the loop you added at least once.

src/ParameterStoreBuildWrapperPlugin.groovy
```diff
@@ -26,6 +32,11 @@ class ParameterStoreBuildWrapperPlugin implements TerraformEnvironmentStagePlugi
 
         stage.decorate(PLAN, addParameterStoreBuildWrapper(options))
         stage.decorate(APPLY, addParameterStoreBuildWrapper(options))
+
+        globalParameterOptions.each { gp ->
+            stage.decorate(PLAN, addParameterStoreBuildWrapper(gp))
+            stage.decorate(APPLY, addParameterStoreBuildWrapper(gp))
+        }
```
kmanning, line 39: Looks like code coverage went down. You modified the apply method, but there are no unit tests exercising that. Can you add a unit test that exercises the new code that you added? The new unit tests should run apply against a Stage, and should have global parameter paths, so that we exercise the loop you added at least once.

bondezbond: I will

src/ParameterStoreBuildWrapperPlugin.groovy
bondezbond: kk captured it here #304

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +26,46 @@ class ParameterStoreBuildWrapperPluginTest {
...
             assertThat(actualPlugins, hasItem(instanceOf(ParameterStoreBuildWrapperPlugin.class)))
         }
+
+        @Test
+        void modifiesTerraformEnvironmentStageCommandWithGlobalParameter() {
+            ParameterStoreBuildWrapperPlugin.withGlobalParameter('/somePath/').init()
+
+            Collection actualPlugins = TerraformEnvironmentStage.getPlugins()
+            assertThat(actualPlugins, hasItem(instanceOf(ParameterStoreBuildWrapperPlugin.class)))
+        }
```
kmanning, line 36: `class Init` is meant to indicate that we're only unit testing the `.init()` method.  The definition of `init()` is 

https://github.com/manheim/terraform-pipeline/blob/6e2272d1fbeda7701d92a4ad480594bf55ef3ba9/src/ParameterStoreBuildWrapperPlugin.groovy#L10-L13

`init()` modifies the TerraformValidateStage, and the TerraformEnvironmentStage.  Those two behaviors are covered by the existing `modifiesTerraformEnvironmentStageCommand` test, and the new `modifiesTerraformValidateStageCommand`.  

I would not mix `withGlobablParameter` into the `class Init`, since `withGlobalParameter` is not called as part of init, and is not expected to be part of `init()`s behavior.  

I would move or remove the above test.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +26,46 @@ class ParameterStoreBuildWrapperPluginTest {
...
+
+        @Test
+        void modifiesTerraformEnvironmentStageCommandWithGlobalParameterAndOptions() {
+            ParameterStoreBuildWrapperPlugin.withGlobalParameter('/somePath/', [someKey: true, anotherKey: 'someValue']).init()
+
+            Collection actualPlugins = TerraformEnvironmentStage.getPlugins()
+            assertThat(actualPlugins, hasItem(instanceOf(ParameterStoreBuildWrapperPlugin.class)))
+        }
```
kmanning, line 44: Same.  We're unit testing `init()`, which does not use `withGlobalParameter`.  I would move or remove this test.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +26,46 @@ class ParameterStoreBuildWrapperPluginTest {
...
+
+        @Test
+        void modifiesTerraformValidateStageCommandWithGlobalParameter() {
+            ParameterStoreBuildWrapperPlugin.withGlobalParameter('/somePath/').init()
+
+            Collection actualPlugins = TerraformValidateStage.getPlugins()
+            assertThat(actualPlugins, hasItem(instanceOf(ParameterStoreBuildWrapperPlugin.class)))
+        }
```
kmanning, line 60: Same for the above.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +26,46 @@ class ParameterStoreBuildWrapperPluginTest {
...
+
+        @Test
+        void modifiesTerraformValidateStageCommandWithGlobalParameterAndOptions() {
+            ParameterStoreBuildWrapperPlugin.withGlobalParameter('/somePath/', [someKey: true, anotherKey: 'someValue']).init()
+
+            Collection actualPlugins = TerraformValidateStage.getPlugins()
+            assertThat(actualPlugins, hasItem(instanceOf(ParameterStoreBuildWrapperPlugin.class)))
+        }
```
kmanning, line 68: Same for the above.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -84,5 +125,38 @@ class ParameterStoreBuildWrapperPluginTest {
             assertEquals(ParameterStoreBuildWrapperPlugin.class, result)
         }
     }
+
+    class withGlobalParameter {
```
kmanning, line 129: I like this.  This helps to indicate that we're unit testing `withGlobalParameter`.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -84,5 +125,38 @@ class ParameterStoreBuildWrapperPluginTest {
...
+        public void reset() {
+            ParameterStoreBuildWrapperPlugin.reset()
+        }
+
+        @Test
+        void addGlobalParameter() {
+            def result = ParameterStoreBuildWrapperPlugin.withGlobalParameter('/path/', [])
+
+            assertEquals([[path: '/path/']], result.globalParameterOptions)
+        }
```
kmanning, line 140: I like this `addGlobalParameter` method.  You're exercising that when `withGlobalParameter` is called, the given paramter is added to the list.  

I might make one modification.  `withGlobalParameter('/path/', [])` <-- is this how you *want* other people to call `withGlobalParameter`, when they have no options?  You've define options to be optional `withGlobalParameter(String path, options=[])`, which means you could call it *without* the empty array.  

Part of the nice effect of unit tests is illustrating how you expect and want others to use your code.  This is an opportunity to illustrate the  convenience of the method you designed.  You  could  call your method  with  `withGlobalParameter('/path/')`

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -84,5 +125,38 @@ class ParameterStoreBuildWrapperPluginTest {
...
+
+        @Test
+        void addGlobalParameterWithOptions() {
+            def result = ParameterStoreBuildWrapperPlugin.withGlobalParameter('/path/', [recursive: true, basename: 'relative'])
+
+            assertEquals([[path: '/path/', recursive: true, basename: 'relative']], result.globalParameterOptions)
+        }
```
kmanning, line 147: I like this test.  It shows that you can call `addGlobalParameters` with or without options.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -84,5 +125,38 @@ class ParameterStoreBuildWrapperPluginTest {
...
+            ArrayList expected = []
+            def result = ParameterStoreBuildWrapperPlugin.withGlobalParameter('/path/', [])
+                                                         .withGlobalParameter('/path2/', [recursive: true])
+                                                         .withGlobalParameter('/path3/', [basename: 'something'])
+            expected << [path:'/path/']
+            expected << [path: '/path2/', recursive: true]
+            expected << [path: '/path3/', basename: 'something']
+            assertEquals(expected, result.globalParameterOptions)
+        }
+    }
```
kmanning, line 160: I like this.  You're showing that you can add more than  one global parameter, and that  they don't overwrite one another.

kmanning:
Code coverage is going down because the `apply` methods were not previously tested, and your change requires that more lines of  code be added to  apply (which therefore increases the number of  untested code lines).

Let's tackle that by first creating contexts for apply in our test suites.  Since we have 2 apply's - one for TerraformValidateStage and another for TerraformEnvironmentStage - let's create an "ApplyForTerraformValidateStage" context, and a "ApplyForTerraformEnvironmentStage" context, so that each method can be unit tested separately

test/ParameterStoreBuildWrapperPluginTest.groovy
bondezbond: removed

test/ParameterStoreBuildWrapperPluginTest.groovy
bondezbond: removed

test/ParameterStoreBuildWrapperPluginTest.groovy
bondezbond: removed

test/ParameterStoreBuildWrapperPluginTest.groovy
bondezbond: removed

test/ParameterStoreBuildWrapperPluginTest.groovy
bondezbond: resolved to show both no options passed and empty options passed as well

bondezbond:
I need a second look at the steps 2-5

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +33,154 @@ class ParameterStoreBuildWrapperPluginTest {
...
+            Map expected        = [path: "/${organization}/${repoName}/${environment}/".toString()]
+
+            configureJenkins(repoName: repoName, organization: organization)
+            ParameterStoreBuildWrapperPlugin plugin = spy(new ParameterStoreBuildWrapperPlugin())
+            doReturn(expected.path.toString()).when(plugin).pathForEnvironment(environment)
+
+            Map actual = plugin.getEnvironmentParameterOptions(environment)
+
+            assertEquals(expected.path, actual.path)
+        }
```
kmanning, line 167: I might consider simplifying this.  

1. `expected` doesn't  have to be a Map.  You're putting an expected path into `expected`, then always pulling path out.  You could make a simpler 1-to-1 implementation with: 
```
        @Test
        void validPathBasedOnEnvironment() {
            String organization = "MyOrg"
            String repoName     = "MyRepo"
            String environment  = "qa"
            Map expectedPath = "/${organization}/${repoName}/${environment}/".toString()

            configureJenkins(repoName: repoName, organization: organization)
            ParameterStoreBuildWrapperPlugin plugin = spy(new ParameterStoreBuildWrapperPlugin())
            doReturn(expectedPath.toString()).when(plugin).pathForEnvironment(environment)

            Map actual = plugin.getEnvironmentParameterOptions(environment)

            assertEquals(expectedPath, actual.path)
        }
```
2. Once you do that, the path itself is irrelevant  to the method under test (getEnvironmentParameterOptions).  Org/Repo - that's not actually known to or used by getEnvironmentParameterOptions.  So  we  could simplify further.
```
        @Test
        void validPathBasedOnEnvironment() {
            String environment  = "qa"
            Map expectedPath = "somepath"
            ParameterStoreBuildWrapperPlugin plugin = spy(new ParameterStoreBuildWrapperPlugin())
            doReturn(expectedPath.toString()).when(plugin).pathForEnvironment(environment)

            Map actual = plugin.getEnvironmentParameterOptions(environment)

            assertEquals(expectedPath, actual.path)
        }
```

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +33,154 @@ class ParameterStoreBuildWrapperPluginTest {
...
+            Map expected        = [credentialsId: "${environment.toUpperCase()}_PARAMETER_STORE_ACCESS"]
+
+            configureJenkins(repoName: repoName, organization: organization)
+            ParameterStoreBuildWrapperPlugin plugin = spy(new ParameterStoreBuildWrapperPlugin())
+            doReturn(expected.path.toString()).when(plugin).pathForEnvironment(environment)
+
+            Map actual = plugin.getEnvironmentParameterOptions(environment)
+
+            assertEquals(expected.credentialsId, actual.credentialsId)
+        }
```
kmanning, line 183: The same simplifications above can be made here.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +33,154 @@ class ParameterStoreBuildWrapperPluginTest {
...
+        @Test
+        void modifiesTerraformValidateStageCommand() {
+            ParameterStoreBuildWrapperPlugin.init()
+
+            Collection actualPlugins = TerraformValidateStage.getPlugins()
+            assertThat(actualPlugins, hasItem(instanceOf(ParameterStoreBuildWrapperPlugin.class)))
+        }
+    }
+
+    public class Apply {
```
kmanning, line 46: The idea behind the classes is to logically group tests.  Junit doesn't let you group tests, so  using classes is a kind of work around.  As a convention, I try to name each grouping/class after the method that I want to test. 

In your case, there are actually 2 `apply` methods.  One that works with TerraformValidateStage, and one that works with TerraformEnvironmentStage.  They're different methods  with different behaviors.  I would consider splitting them out.  Using the TagPlugin  as an example, you could do an "ApplyForTerraformValidateStage", and an "ApplyForTerraformEnvironmentStage".  Or you could even do nested:
```
...
    class Apply {
        class WithTerraformValidateStage {
             ....
        }

        class  WithTerraformEnvironmentStage {
            ....
        }
    }
...
```

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +33,154 @@ class ParameterStoreBuildWrapperPluginTest {
...
+            ParameterStoreBuildWrapperPlugin plugin = spy(new ParameterStoreBuildWrapperPlugin())
+
+            doReturn(environment).when(stage).getEnvironment()
+            doReturn(apo).when(plugin).getEnvironmentParameterOptions(environment)
+            doReturn(expectedClosure).when(plugin).addParameterStoreBuildWrapper(apo)
+
+            plugin.apply(stage)
+
+            verify(stage, times(2)).decorate(anyString(), eq(expectedClosure))
+        }
```
kmanning, line 107: The intent behind this test is a little confusing, and I think  the  problem is being  made more difficult with the extra setup that's not relevant to testing the `apply` method?

test/ParameterStoreBuildWrapperPluginTest.groovy
bondezbond: done

test/ParameterStoreBuildWrapperPluginTest.groovy
bondezbond: done

test/ParameterStoreBuildWrapperPluginTest.groovy
bondezbond: opted for the nested route

test/ParameterStoreBuildWrapperPluginTest.groovy
kmanning: It's worth nothing, that even with these changes to simplify, your previous tests have uncovered another "smell" that existed before you even picked up this PR.  Does `addParameterStoreBuildWrapper` really need to be called twice?

```
        def environment = stage.getEnvironment()	        
        def options << getParameterOptions(environment)

        options.each { option ->
            def buildWrapperClosure = addParameterStoreBuildWrapper(option)
            stage.decorate(PLAN, buildWrapperClosure)
            stage.decorate(APPLY, buildWrapperClosure)
        }
```

Now your test would have even less to validate.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +33,236 @@ class ParameterStoreBuildWrapperPluginTest {
...
+            ParameterStoreBuildWrapperPlugin.reset()
+        }
+
+        private configureJenkins(Map config = [:]) {
+            Jenkinsfile.instance = mock(Jenkinsfile.class)
+            when(Jenkinsfile.instance.getStandardizedRepoSlug()).thenReturn(config.repoSlug)
+            when(Jenkinsfile.instance.getRepoName()).thenReturn(config.repoName ?: 'repo')
+            when(Jenkinsfile.instance.getOrganization()).thenReturn(config.organization ?: 'org')
+            when(Jenkinsfile.instance.getEnv()).thenReturn(config.env ?: [:])
+        }
```
kmanning, line 59: Is this method being used?  I think we pushed out the need for it?

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +33,236 @@ class ParameterStoreBuildWrapperPluginTest {
...
```
kmanning, line 160: Same here - is this being used?  I think we pushed out the need for it.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +33,236 @@ class ParameterStoreBuildWrapperPluginTest {
...
+            Map option                              = [:]
+            List expected                           = [ option ]
+            ParameterStoreBuildWrapperPlugin plugin = spy(new ParameterStoreBuildWrapperPlugin())
+
+            doReturn(option).when(plugin).getEnvironmentParameterOptions(environment)
+
+            List actual = plugin.getParameterOptions(environment)
+
+            assertEquals(expected, actual)
+        }
```
kmanning, line 174: `returnsNoOptionsWhenNotSet` that name is weird, and doesn't match with what's being asserted. I think this is the result of reshuffling  behavior.  It's not that "NoOptions" are returned - it's that only environment-specific options are returned.

test/ParameterStoreBuildWrapperPluginTest.groovy
```diff
@@ -25,6 +33,236 @@ class ParameterStoreBuildWrapperPluginTest {
...
+            Map option                              = [ key: "value" ]
+            List expected                           = [ option ]
+            ParameterStoreBuildWrapperPlugin plugin = spy(new ParameterStoreBuildWrapperPlugin())
+
+            doReturn(option).when(plugin).getEnvironmentParameterOptions(environment)
+
+            List actual = plugin.getParameterOptions(environment)
+
+            assertEquals(expected, actual)
+        }
```
kmanning, line 188: This seems like a duplicate of the  test  above.

kmanning:
Awesome!  Thanks for the contribution, this feels like a super useful feature!  :shipit:

Issue #311, opened 2020-12-09
Improper null checking leads to test failures


jantman:
Running the ``./gradlew check --info`` tests locally on my system resulted in four failed tests:

```
FileParametersPluginTest > FileParametersPluginTest$GetVariables.handlesCarriageReturnCharacters FAILED
    java.lang.NullPointerException: Cannot get property 'env' on null object
        at org.codehaus.groovy.runtime.NullObject.getProperty(NullObject.java:60)
        at org.codehaus.groovy.runtime.InvokerHelper.getProperty(InvokerHelper.java:174)
        at org.codehaus.groovy.runtime.callsite.NullCallSite.getProperty(NullCallSite.java:47)
        at org.codehaus.groovy.runtime.callsite.GetEffectivePogoFieldSite.callGetProperty(GetEffectivePogoFieldSite.java:42)
        at Jenkinsfile.getEnv(Jenkinsfile.groovy:113)
        at Jenkinsfile$getEnv.call(Unknown Source)
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:117)
        at FileParametersPlugin.getEnv(FileParametersPlugin.groovy:45)
        at FileParametersPlugin.interpolate(FileParametersPlugin.groovy:41)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:93)
        at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
        at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:384)
        at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1022)
[...]

PR #313, opened 2020-12-09
Issue #311: non-deterministic test failures

kmanning:
Issue #311 
* Mockito spies use different stub syntax.  `getEnv` should be stubbed, but is getting incorrectly called instead.  Use the correct stubbing syntax so that the underlying method is not called.
* Identify and fix the test that was not cleaning up after itself (Answer: TagPluginTest)
* Fix FileParametersPlugin - it's accessing global environment variables, and therefore expects `Jenkinsfile.original` to be stubbed.


jantman (approved):
This all looks good to me, thanks!

PR #317, opened 2020-12-11
Fixes #316 - implement more granular decorations in TerraformEnvironmentStage

### LINKED ISSUE #316: Allow more granular decorations around terraform init/plan/apply
This is a subset of #314. I would like to allow more granular decoration around individual terraform init/plan/apply commands in TerraformEnvironmentStage, to enable a future shell hook plugin.

jantman:
This PR adds some more-granular decorators around the shell commands run in TerraformEnvironmentStage (init, plan, and apply). This provides an extension point for a future plugin to allow executing shell commands (hooks) before and/or after each of these commands.

Please note that I have not (yet?) added unit tests for this. As far as I can tell the current code does not have tests for anything in the closure, aside from it not blowing up, and I'm not entirely sure how to test this...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * [Issue #311](https://github.com/manheim/terraform-pipeline/issues/311) Fix non-deterministic test failures
+* [Issue #316](https://github.com/manheim/terraform-pipeline/issues/316) Implement more granular decorations in TerraformEnvironmentStage.
 
 # v5.12
 
--- a/docs/TerraformEnvironmentStage.md
+++ b/docs/TerraformEnvironmentStage.md
@@ -42,3 +42,9 @@ validate.then(deployQA)
         .then(deployProd)
         .build()
 ```
+
+### For Plugin Developers
+
+#### Decorate Stages/Commands or Hook before/after command execution
+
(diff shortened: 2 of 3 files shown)


kmanning:
Being able to inject behavior with Closures is delegated down to the StageDecoration class.  And that behavior is exercised - https://github.com/manheim/terraform-pipeline/blob/master/test/StageDecorationsTest.groovy

Initially, the `pipelineConfiguration` methods were not tested at all, because that's the section that integrates with Jenkinsfile DSL.  A method that integrates with a lot of external classes is difficult to test in a unit test.  That said, since this library was initially published, a DummyJenkinsfile was since added, which now allows you to stub out and assert on DSL methods.

To directly answer your question about how to test this - we could try to spy the StageDecorations, and could assert that `apply` is called for the new points of code which can now be decorated.

kmanning:
Thoughts on whether or not `INIT_COMMAND` for plan should be separate from `INIT_COMMAND` for apply?

jantman:
Ok, that makes sense about testing. Thanks.

> Thoughts on whether or not `INIT_COMMAND` for plan should be separate from `INIT_COMMAND` for apply?

Personally I don't think so. In fact, I think that specifically shouldn't be done, because it would make it easier to do the wrong thing, and have undesirable differences between the plan and the apply. By keeping the `INIT_COMMAND` the same across all init's, we make it easy to add in something that should be done everywhere `terraform init` is run.

kmanning:
Looks good, merging.

Issue #250, opened 2020-07-16
Provide a hook for arbitrary scripts after 'apply'


kmanning:
UseCase: Healthcheck
* I've deployed infrastructure to an environment, using terraform-pipeline
* I want to confirm that the infrastructure "works as expected"
* If I had a "hook" that allowed me to run an arbitrary shell script, my shell script could validate the my infrastructure "works as expected"
* If my shell script returns a zero-status, I assume "it works" (based on unix return code conventions)
* If my shell script returns a non-zero status, I assume that "it did not work as expected", and marks the "apply" stage of my environment as failed.  I'm not  allowed to continue.
* My shell script should get all the "goodness" of pipelines that operate during "apply".  Eg: I can access the IAM role and EnvironmentVariables provided by plugins like ParameterStorePlugin.

kmanning:
UseCase: Database migrations
* I've deployed a database using terraform-pipeline
* I'd like to run a database migration, to ensure that my database schema is altered in an automated fashion, and all my environments are kept in-sync.
* If my shell script returns a zero-status, I assume "it works" (based on unix return code conventions)
* If my shell script returns a non-zero status, I assume that "it did not work as expected", and marks the "apply" stage of my environment as failed. I'm not allowed to continue.
* My shell script should get all the "goodness" of pipelines that operate during "apply". Eg: I can access the IAM role and EnvironmentVariables provided by plugins like ParameterStorePlugin.

kmanning:
Other thoughts: 
- This could potentially be a hook for ChangeMangement tasks

kmanning:
HealthCheck Script
```
@Library(['terraform-pipeline@v5.1']) _

Jenkinsfile.init(this)
PostApplyPlugin.run('./bin/healthcheck.sh')

def validate = new TerraformValidateStage()
def deployQa = new TerraformEnvironmentStage('qa')
def deployUat = new TerraformEnvironmentStage('uat')
def deployProd = new TerraformEnvironmentStage('prod')

validate.then(deployQa)
        .then(deployUat)
        .then(deployProd)
        .build()
```

DatabaseMigrate Script
```
@Library(['terraform-pipeline@v5.1']) _

Jenkinsfile.init(this)
PostApplyPlugin.run('./bin/run_migrations.sh')

def validate = new TerraformValidateStage()
def deployQa = new TerraformEnvironmentStage('qa')
def deployUat = new TerraformEnvironmentStage('uat')
[...]

PR #319, opened 2020-12-18
Issues/250 - Implement shell hook plugin

jantman:
This PR is an initial, proof-of-concept implementation of a shell script hook plugin for TerraformEnvironmentStage, as described in #250. The plugin name (``TerraformEnvironmentStageShellHookPlugin``) is intended to be an indication that this is _not_ an attempt at an all-encompassing solution to the hook problem, and is one initial implementation that may or may not inspire something at a higher level.

The code included here has full unit test coverage, and has also been tested in a Jenkinsfile on a private Jenkins instance. All seems to be working well.


jantman:
The [TravisCI tests](https://travis-ci.com/github/manheim/terraform-pipeline/builds/209643807) are failing with:

```
Execution failed for task ':compileTestGroovy'.

> BUG! exception in phase 'class generation' in source unit '/home/travis/build/manheim/terraform-pipeline/test/HookPointTest.groovy' unsupported Target MODULE
```

I have no idea what's causing this. ``./gradlew check --info`` succeeds locally for me in my working clone. In a fresh clone, it seems to be failing with an unrelated order-dependent error ( looks like a resurgence of the same bug that #313 sought to fix).

docs/TerraformEnvironmentStageShellHookPlugin.md
```diff
@@ -0,0 +1,44 @@
+## [TerraformEnvironmentStageShellHookPlugin](../src/TerraformEnvironmentStageShellHookPlugin.groovy)
+
+**Note:** This plugin is intended as a limited proof-of-concept of a hook system. It (and its unweildy long name) is intended to be replaced at some point in the future with a more flexible and generic, non-TerraformEnvironmentStage-specific hook system.
```
kmanning, line 3: Honestly, this makes me a little uncomfortable.  I don't like the idea of publishing a plugin that's intended to be a proof-of-concept.  If that's the case, we could publish this plugin in a separate library, learn from it, and then merge the results of what we learn later.

Thoughts?

src/TerraformEnvironmentStageShellHookPlugin.groovy
```diff
@@ -0,0 +1,60 @@
...
+
+class TerraformEnvironmentStageShellHookPlugin implements TerraformEnvironmentStagePlugin {
+    static hooks = [
+        (ALL): new HookPoint(ALL),
+        (INIT_COMMAND): new HookPoint(INIT_COMMAND),
+        (PLAN): new HookPoint(PLAN),
+        (PLAN_COMMAND): new HookPoint(PLAN_COMMAND),
+        (APPLY): new HookPoint(APPLY),
+        (APPLY_COMMAND): new HookPoint(APPLY_COMMAND)
+    ]
```
kmanning, line 16: Would it be easier if this here initialized as an empty map?

docs/TerraformEnvironmentStageShellHookPlugin.md
```diff
@@ -0,0 +1,44 @@
...
+def deployUat = new TerraformEnvironmentStage('uat')
+def deployProd = new TerraformEnvironmentStage('prod')
+
+validate.then(deployQA)
+        .then(deployUat)
+        .then(deployProd)
+        .build()
+```
+
+**Reminder:** Please remember that plugin definition order matters. You most likely want this plugin to be defined in your Jenkinsfile _before_ any plugins that affect the environment.
```
kmanning, line 44: Thanks for pointing this out.  It's worth noting that with this implementation, all hooks have to follow the same order, relative to other plugins.  (vs, an implementation where the different hook points were treated as individual Plugins, and could therefore be ordered independently of one another).  <-- I'm only saying this in the spirit of a proof-of-concept.

kmanning:
> The [TravisCI tests](https://travis-ci.com/github/manheim/terraform-pipeline/builds/209643807) are failing with:
> 
> ```
> Execution failed for task ':compileTestGroovy'.
> 
> > BUG! exception in phase 'class generation' in source unit '/home/travis/build/manheim/terraform-pipeline/test/HookPointTest.groovy' unsupported Target MODULE
> ```
> 
> I have no idea what's causing this. `./gradlew check --info` succeeds locally for me in my working clone. In a fresh clone, it seems to be failing with an unrelated order-dependent error ( looks like a resurgence of the same bug that #313 sought to fix).

I don't think that's related to issue #313.  

I can't reproduce locally, but I did find the same error referenced elsewhere.  
 The thread seems to suggest it's a problem with a mismatch between java/gradle versions. 
 It's unclear why your PR would trigger this - master continues to run find with the same combination. 

 https://stackoverflow.com/questions/45899027/errorbug-exception-in-phase-class-generation-in-source-unit-buildscript

kmanning:
Looks like JobDSL 1.77 upgraded groovy to 2.4.12.  The CHANGELOG for 2.4.12 notes a fix for compile errors in legacy version. 

* http://groovy-lang.org/changelogs/changelog-2.4.12.html
* https://issues.apache.org/jira/browse/GROOVY-8235

Upgrading groovy seems to make the problem go away.  Upgrading gradle reveals the problem locally (which doesn't make sense, since travisCI is using ./gradlew - same as what we'd be using locally).  

At any rate, I'll cut a difficult issue to address this.

kmanning:
PR for Issue #321 is available (See #322) - I can merge the change into master, and you can rebase to fix your PR.

kmanning:
Also, can you update the CHANGELOG for this PR?

docs/TerraformEnvironmentStageShellHookPlugin.md
jantman: Do we have a place to list such libraries? Do you think that we would actually learn anything that way?

In general, my thoughts are:

1. I think this plugin provides value to users other than myself. I'm not crazy about putting it somewhere that's unlikely for anyone else to find it.
2. I think learning from it will only come with visibility and use. *I* already have an idea of how I think a hook system would work... and this plugin is an example of it.

In general, the process that I was thinking of is that this plugin would be published as-is, and one of the following would eventually happen:

1. We implement a generic hook system that's compatible with this plugin, and it becomes... no longer a proof of concept. That note gets removed.
2. We implement a generic hook system that's not compatible with this plugin, and we deprecate this plugin. We provide migration instructions in release notes, and give users a few versions to switch before we remove this plugin.
3. We don't (at least in the near future) implement a generic hook system, and this plugin lives on until something changes.

src/TerraformEnvironmentStageShellHookPlugin.groovy
jantman: On the flip side, it means that ``withHook()`` would need to null-check each of the values, and construct the new HookPoint objects if they're needed. Maybe this is backwards, but I wanted to contain all of the logic of "is this hook set" inside the HookPoint class, which led to the current implementation. Though I'm certainly open to changing it if you prefer.

docs/TerraformEnvironmentStageShellHookPlugin.md
jantman: Yeah, that's certainly true. And my view on this might be heavily biased towards my own use case... which is essentially that all of the hooks should be run as similarly to the terraform commands as possible (i.e. with the same wrappers).

jantman:
@kmanning I've updated the changelog as requested. Once #322 is merged to master I'll rebase. Still have a few discussion items pending from your review.

kmanning:
Merged

jantman:
@kmanning Rebased and tests passing. Thanks!!

docs/TerraformEnvironmentStageShellHookPlugin.md
kmanning: To rephrase - I'm uncomfortable with the idea of embedding "this is a proof of concept" in our docs, and not understanding who's going to manage that next step, or worse, we don't take the next step.  Are you accepting that responsibility? If not, and if I will be responsible, I'd prefer to either publish this and say "this is complete (enough)" and treat it as such, or to publish it in a separate library until we decide that it's complete enough.

2. "learning from it will only come with visibility and use" - Agreed.  We've got an internal library for Cox-specific plugins - I'd suggest putting it there, so that we could use it company wide and learn.  I feel like our organization is more than large enough that we'd get enough use/feedback to learn whether or not it's complete enough, and decide where to go from there.

1. "We implement a generic hook system that's compatible with this plugin, and it becomes... no longer a proof of concept. That note gets removed." - who's job will it be to do this?
3. "We don't (at least in the near future) implement a generic hook system, and this plugin lives on until something changes." - that's my discomfort. I don't want that "this is a proof of concept" to live on indefinitely.

jantman: I think it's complete enough to live on indefinitely as a hook system that's specific to wrapping TES commands in shell hooks. However, I'm not sure what happens to it if/when we get a more complete, generic hook system. That was my main concern around "proof of concept"... that at some point in the future, some unknown person would implement a generic hook system that may or may not be compatible with this.

How do you feel about me just removing this note, and seeing what the future brings?

src/TerraformEnvironmentStageShellHookPlugin.groovy
kmanning: To be more specific, I think there's a lot of examples like this, where you're iterating over all the hooks because they're all non-null.  Looping can be tricky in a test - there's actually no guarantee that a PLAN_COMMAND entry will even exist in the list.

```
        @Test
        void setRunBefore() {
            TerraformEnvironmentStageShellHookPlugin.withHook(PLAN_COMMAND, 'foo bar', WhenToRun.BEFORE)
            def hooks = TerraformEnvironmentStageShellHookPlugin.hooks
            hookKeys.each {
                if (it == PLAN_COMMAND) {
                    assertTrue(hooks[it].isConfigured())
                    assertEquals(hooks[it].runBefore, 'foo bar')
                    assertNull(hooks[it].runAfterAlways)
                    assertNull(hooks[it].runAfterOnFailure)
                    assertNull(hooks[it].runAfterOnSuccess)
                } else {
                    assertFalse(hooks[it].isConfigured())
                }
            }
        }
```

I was wondering if it might be more straight forward to do something like this, which obviates the need to loop, and by passes the problem of the element not being in the list.

```
        @Test
        void setRunBefore() {
            TerraformEnvironmentStageShellHookPlugin.withHook(PLAN_COMMAND, 'foo bar', WhenToRun.BEFORE)
            def hooks = TerraformEnvironmentStageShellHookPlugin.hooks
            assertEquals(1, hooks.size())
            
            def planHooks = hooks[PLAN_COMMAND]
            assertTrue(planHooks.isConfigured())
            assertEquals(planHooks.runBefore, 'foo bar')
            assertNull(planHooks.runAfterAlways)
            assertNull(planHooks.runAfterOnFailure)
            assertNull(planHooks.runAfterOnSuccess)
        }
```

Which then suggests even more simplification with:

```
        @Test
        void setRunBefore() {
            TerraformEnvironmentStageShellHookPlugin.withHook(PLAN_COMMAND, 'foo bar', WhenToRun.BEFORE)
            def planHooks = TerraformEnvironmentStageShellHookPlugin.getHooks(PLAN_COMMAND)
            assertTrue(planHooks.isConfigured())
            assertEquals(planHooks.runBefore, 'foo bar')
            assertNull(planHooks.runAfterAlways)
            assertNull(planHooks.runAfterOnFailure)
            assertNull(planHooks.runAfterOnSuccess)
        }
```

It's a suggestion though, just based on looking at the tests, and not looking at the implementation.

jantman: So... those alternatives are missing the intended point of the test, or half of it.

```
        @Test
        void setRunBefore() {
            TerraformEnvironmentStageShellHookPlugin.withHook(PLAN_COMMAND, 'foo bar', WhenToRun.BEFORE)
            def hooks = TerraformEnvironmentStageShellHookPlugin.hooks
            hookKeys.each {
                if (it == PLAN_COMMAND) {
                    assertTrue(hooks[it].isConfigured())
                    assertEquals(hooks[it].runBefore, 'foo bar')
                    assertNull(hooks[it].runAfterAlways)
                    assertNull(hooks[it].runAfterOnFailure)
                    assertNull(hooks[it].runAfterOnSuccess)
                } else {
                    assertFalse(hooks[it].isConfigured())
                }
            }
        }
```

Is intended to test that the plan before hook, and ONLY the plan before hook, has been set. The iteration is there to ensure that none of the other hooks, even though they're present, have been touched. If there was an easy way to serialize the complete state of ``hooks`` and test against that, then I'd do it.

The iteration here was just intended to simplify:

```
assertFalse(TerraformEnvironmentStageShellHookPlugin.getHooks(INIT_COMMAND).isConfigured())
assertFalse(TerraformEnvironmentStageShellHookPlugin.getHooks(PLAN).isConfigured())
def planHooks = TerraformEnvironmentStageShellHookPlugin.getHooks(PLAN_COMMAND)
assertTrue(planHooks.isConfigured())
assertEquals(planHooks.runBefore, 'foo bar')
assertNull(planHooks.runAfterAlways)
assertNull(planHooks.runAfterOnFailure)
assertNull(planHooks.runAfterOnSuccess)
assertFalse(TerraformEnvironmentStageShellHookPlugin.getHooks(APPLY).isConfigured())
assertFalse(TerraformEnvironmentStageShellHookPlugin.getHooks(APPLY_COMMAND).isConfigured())
```

On another note:

> there's actually no guarantee that a PLAN_COMMAND entry will even exist in the list.

Why not? Am I missing something?

docs/TerraformEnvironmentStageShellHookPlugin.md
kmanning: I vote removing the note, and seeing what the future brings.

src/TerraformEnvironmentStageShellHookPlugin.groovy
kmanning: > The iteration is there to ensure that none of the other hooks, even though they're present, have been touched

That's where I was suggesting that using nulls might be helpful to simply that intent.  

```
            def hooks = TerraformEnvironmentStageShellHookPlugin.hooks
            assertEquals(1, hooks.size())
            ...
            def planHooks = hooks[PLAN_COMMAND]
            ...
```

The existence of a single non-null hook confirms that all other hooks are not set.   No worries though, it was just a suggestion. 

> Is intended to test that the plan before hook, and ONLY the plan before hook, has been set.
> On another note:
>  > there's actually no guarantee that a PLAN_COMMAND entry will even exist in the list.
>
> Why not? Am I missing something?

Apparently, I missed something, and not you.  I didn't realize until now that you created a `hookKeys` array, and are iterating over **THAT**, and not iterating over the hooks - so I'm wrong.

```
            def hooks = TerraformEnvironmentStageShellHookPlugin.hooks
       -->  hookKeys.each {
                if (it == PLAN_COMMAND) {
                    assertTrue(hooks[it].isConfigured())
                    assertEquals(hooks[it].runBefore, 'foo bar')
                    assertNull(hooks[it].runAfterAlways)
                    assertNull(hooks[it].runAfterOnFailure)
                    assertNull(hooks[it].runAfterOnSuccess)
                } else {
                    assertFalse(hooks[it].isConfigured())
                }
            }
```

docs/TerraformEnvironmentStageShellHookPlugin.md
jantman: Sounds good to me. I've pushed up a commit that removes the PoC note.

Issue #172, opened 2019-12-12
ConditionalApplyPlugin: Allow 'apply' on branches for specific environments


boommies:
**Use Case:**
I want to be able to apply changes to non-production environments, without merging to master.

Example: I have 3  environments: 
* qa
* uat
* prod

I want to be able to make a feature branch in git, and apply the changes from that branch to qa, before merging it into uat and prod.

**Example Code:** 
```
@Library(['terraform-pipeline']) _

Jenkinsfile.init(this)

ConditionalApplyPlugin.withBranchApplyEnabledFor('qa').init()

def validate = new TerraformValidateStage()
def deployQa = new TerraformEnvironmentStage('qa')
def deployUat = new TerraformEnvironmentStage('uat')
def deployProd = new TerraformEnvironmentStage('prod')

validate.then(deployQa)
        .then(deployUat)
        .then(deployProd)
        .build()
```

**Expected:** 

When I create a git branch and push a change, the pipeline will:
1. Run a terraform plan for  'qa'
2. Allow the user to apply the plan to 'qa'
3. Run a plan for 'uat' (*DOES NOT* allow apply for 'uat')
4. Run a plan for 'prod' (*DOES NOT* allow apply for 'uat')

boommies:
Alternative to above:

When I create a git branch and push a change, the pipeline will:

1. Run a plan for 'qa'
2. Run a plan for 'uat'
3. Run a plan for 'prod'
4. Allow the user to apply the plan for 'qa' (only at the end)

ChrisBubelisCAI:
I like the first approach, with the subtle change that it is driven by the creation of a pull request and commits to that pull request branch.

kmanning:
In hindsight, I think option 1 is the most easily implemented as well.  Cool, let's set that as the goal for this issue.

kmanning:
Bumping to a future release

kmanning:
Bumping to a future release

kmanning:
bumping to a future release

kmanning:
Related: https://github.com/manheim/terraform-pipeline/issues/328 - this issue would allow `apply` across all environments on some branches.

kmanning:
Given the similarity between issue #172 and #328, it feels like we should change the suggested method names in the example code above.  What if: 

For Issue #172: apply on branches for some environments
```
.withApplyOnEnvironment('qa')
```

For Issue #328: 
```
.withApplyOnBranch('master')
```

[...]

Issue #25, opened 2019-09-03
ConditionalApplyPlugin - allow apply on specific branches


kmanning:
* Allow the branch (master) to be customized
* (REMOVING disable - See Issue #329 )

kemper:
For allowing this plugin to be disabled, I take that to mean always apply? I can update my changes to support that. One option is to take an empty list of branches to mean "all"

```
ConditionalApplyPlugin.withBranchApplyEnabledFor([])
```

or an extra option:

```
ConditionalApplyPlugin.withBranchApplyEnabledFor([], all: true)
```

or a wildcard:

```
ConditionalApplyPlugin.withBranchApplyEnabledFor(['*'])
```

or I can add a new method and new field for controlling enable/disable.

```
ConditionalApplyPlugin.setEnabled(false)
```

kmanning:
Let's split  this issue into 2 - they are distinct.   I  updated the  original post

kmanning:
Issues #25 and #172 are related.  I wonder if the code examples should be changed to reflect the relationship.  Consider the following: 

Issue #25 - allow apply on specific *branches*
```
ConditionalApplyPlugin.withApplyOnBranch('master')
```

Issue #172 - allow apply on specific *branches*
```
ConditionalApplyPlugin.withApplyOnEnvironment('qa')
```

kemper:
Sounds good to me. Current the name/parameter combination on the PR is `withBranchApplyEnabledFor(['branch1', 'branch2')` . Do you want me to rename and change the params to be varargs?

kmanning:
Yes on the rename.

On the varargs, it feels friendlier.  So yes, I'd personally appreciate it, unless there's a good reason not to.

PR #325, opened 2021-01-07
Be able to configure conditional branches

kemper:
resolves manheim/terraform-pipeline#172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* [Issue #321](https://github.com/manheim/terraform-pipeline/issues/172) Allow 'apply' on branches other than master for some environments
+
 # v5.13
 
 * [Issue #321](https://github.com/manheim/terraform-pipeline/issues/321) Upgrade groovy from 2.4.11 to 2.4.12 (fix travisCi failures)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* [Issue #321](https://github.com/manheim/terraform-pipeline/issues/172) Allow 'apply' on branches other than master for some environments
+* [Issue #25](https://github.com/manheim/terraform-pipeline/issues/25) Allow 'apply' on branches other than master for some environments
 
 # v5.13
 
--- a/src/ConditionalApplyPlugin.groovy
+++ b/src/ConditionalApplyPlugin.groovy
@@ -3,10 +3,15 @@ import static TerraformEnvironmentStage.APPLY
 
 public class ConditionalApplyPlugin implements TerraformEnvironmentStagePlugin {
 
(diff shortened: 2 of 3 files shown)


kemper:
I tried to follow the initial suggestion on issue #172 e.g. `ConditionalApplyPlugin.withBranchApplyEnabledFor('qa').init()` but I'm new to the terraform-pipeline project and didn't understand the intent of calling `.init()` at the end. I figured I'd take a stab at it and get feedback afterwards.

kmanning:
I wanted to design the library in a way that allowed plugins to be enabled or disabled, as the user sees fit.  I also wanted the enable/disable to be done as simply as possible - a static method fit the bill.  

So `init()` is the convention that I came up with, to enable a plugin - without calling `init()`, a plugin should not have any effect on your pipeline, with the exception of a handful of "default" plugins (which includes ConditionalApplyPlugin) See: https://github.com/manheim/terraform-pipeline/blob/affa2f11c59bde1c388209b681b37d64327835a9/src/TerraformEnvironmentStage.groovy#L7

This is definitely an awkward implementation in hindsight.  A better implementation could have treated ConditionalApplyPlugin (and every default plugin), like any other plugin, and had it implement an `init()` method.  That `init()` method could then have been called in `Jenkinsfile.init()`, and achieved the same current behavior, without the code being any different from any other plugin.

kmanning:
The implementation looks good.  Can you update the CHANGELOG with your Issue?  I'm also about to cut a release on master, and this change can go into v5.14, the next release.

kmanning:
Hrm, I think I spent more time thinking/talking about the `init` method, that I didn't look into the implementation details enough.  I think what you implemented is different from the intent outlined in the issue.

Your change will allow `apply` to run on *all environments* so long as the branch names match.

The intent described in the issue is that `apply` can run on *a specific environment*, regardless of what the branch is named.

Those are two different things.  I pointed you in the direction of the existing Issue - is your UseCase different from what was outlined?  If so, we should open a different issue.

kmanning:
Per the discussion about the awkwardness of `init` and `ConditionalApplyPlugin`, I put in Issue #327

kemper:
After chatting with @kmanning I realize I misinterpreted the goal of this issue. My PR does address half of issue #25 however, so I'm planning to update the PR for that issue.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* [Issue #321](https://github.com/manheim/terraform-pipeline/issues/172) Allow 'apply' on branches other than master for some environments
```
kmanning, line 3: Can the Issue number and link be corrected?  Both should point to Issue #25 - https://github.com/manheim/terraform-pipeline/issues/25

kmanning:
Other than the links in the ChangeLog, the changes look good.

Issue #320, opened 2020-12-18
Defect: GithubPRPlanPlugin - failing plan marked as success?


kmanning:
* Investigate this

CliffJumper:
Not sure if this is the same, but with v5.14, I'm seeing terraform plan errors "swallowed".  The net effect is that I don't see the plan output in the Jenkins log.

kmanning:
How to reproduce: 
1. Create a terraform resource that will fail on `terraform plan`.  For example, read from a file that does not exist.
2. Enable the GithubPRPlanPlugin in your pipeline - this will pipeline the output to `plan.out`, and stderr to `plan.err`
3. Run your pipeline
4. Expected: The pipeline should fail at the end of the plan command, and the error message from terraform plan should be displayed.
5. Actual: The pipeline continues.  The stderr output goes unused in the `plan.err` file

PR #338, opened 2021-02-10
Issue 320: Fix bug in GithubPRPlugin, stop pipeline and display error on plan failures.

kmanning:
* Issue #320 
* The Plugin incorrectly swallows plan errors (due to an always-successful `tee` command).
* Fail the pipeline on plan errors, and display the plan results to the user.
* Continue to behave normally if there are no plan errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+* [Issue #320](https://github.com/manheim/terraform-pipeline/issues/320) Bug Fix: GithubPRPlugin should stop on plan errors, and display error to the user.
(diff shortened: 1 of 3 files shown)


jantman (approved):
This looks good to me; as far as I can tell, it should solve the issue.

For what it's worth, the STDOUT of terraform 0.14 gets *very* noisy on large terraform configurations. For my team's own internal library (infrastructure-pipeline) we switched from using this method of using stdout/stderr redirection, to using `terraform show tfplan` for the PR comment.

kmanning:
Cutting a separate issue to improve how the comment is generated.  See Issue #340

Issue #347, opened 2021-02-24
TerraformOutputOnlyPlugin: Add new plugin to limit the pipeline to `terraform output` only.


duckpuppy:
As a user, I would like a way to instruct the pipeline to simply show the outputs existing in the current state without running a new plan or apply.

This new plugin provides the ability to perform the above by providing a set of job parameters that control this behavior.

----

## TerraformOutputOnlyPlugin

Enable this plugin to change pipeline functionality. This plugin will skip the plan and apply stages and add three new job parameters.

* `SHOW_OUTPUTS_ONLY`: This configures the job to skip execution of the plan and apply terraform commands. The job will perform the INIT stage and immediately perform `terraform output`. Unless this option is checked, the following options will have no effect.
* `JSON_FORMAT_OUTPUTS`: This will instruct the plugin to display the output in JSON format.
* `REDIRECT_OUTPUTS_TO_FILE`: Text entered into this option will be used to redirect the result of `terraform output` to a file in the current workspace. The filename should be relative to the workspace, and directories will NOT be created so they should exist beforehand.

```
// Jenkinsfile
@Library(['terraform-pipeline@v3.10']) _

Jenkinsfile.init(this, env)

// This enables the "output only" functionality
// Note that the functionality will NOT be applied unless the `SHOW_OUTPUTS_ONLY` option is `true`.
TerraformOutputOnlyPlugin.init()

def validate = new TerraformValidateStage()

def destroyQa = new TerraformEnvironmentStage('qa')
def destroyUat = new TerraformEnvironmentStage('uat')
def destroyProd = new TerraformEnvironmentStage('prod')

validate.then(destroyQa)
        .then(destroyUat)
        .then(destroyProd)
        .build()
```

Issue #152, opened 2019-09-03
Simplify Plugin Architecture


kmanning:
* The plugins have a clear interface, but it may be overly specific.
* Part of the specificity is the `decorate` method.
* Consider defaulting the `ALL` argument of `decorate` - if you call decorate and only pass it a closure, it should default to ALL
* Consider rewriting some plugins to use `ALL` unless it's absolutely necessary.
* Consider pushing `decorate` into a the `Stage` interface.
* If we did the above, I think the plugin implementation could be a lot more simple and flexible.

PR #348, opened 2021-02-24
Issue 347: Add TerraformOutputOnlyPlugin

duckpuppy:
Fixes issue #347: Create `TerraformOutputOnlyPlugin` to provide the option of producing the current Terraform state outputs, skipping `terraform plan` and `terraform apply`


duckpuppy:
Forgot to mark this as a draft. Haven't committed tests yet.

duckpuppy:
OK, this is not WIP anymore.

src/PassPlanFilePlugin.groovy
```diff
@@ -13,8 +13,8 @@ class PassPlanFilePlugin implements TerraformPlanCommandPlugin, TerraformApplyCo
 
     @Override
     public void apply(TerraformEnvironmentStage stage) {
-        stage.decorate(PLAN,  stashPlan(stage.getEnvironment()))
-        stage.decorate(APPLY, unstashPlan(stage.getEnvironment()))
+        stage.decorate(PLAN_COMMAND,  stashPlan(stage.getEnvironment()))
+        stage.decorate(APPLY_COMMAND, unstashPlan(stage.getEnvironment()))
```
kmanning, line 17: Can you clarify why the stash/unstash needed to be narrowed in this way?   How is it related to outputting?

test/TerraformOutputOnlyPluginTest.groovy
```diff
@@ -0,0 +1,142 @@
...
+                defaultValue: false,
+                description: "Render 'terraform output' results as JSON. Only applies if SHOW_OUTPUTS_ONLY is selected."
+            ]))
+
+            assertThat(actualParms, hasItem([
+                $class: 'hudson.model.StringParameterDefinition',
+                name: "REDIRECT_OUTPUTS_TO_FILE",
+                defaultValue: "",
+                description: "Filename relative to the current workspace to redirect output to. Only applies if 'SHOW_OUTPUTS_ONLY' is selected."
+            ]))
```
kmanning, line 54: I might consider breaking this test case up into 3 separate test cases, and asserting on each parameter individually.  But that's 100% optional.

kmanning (commented):
Looks great overall.

src/PassPlanFilePlugin.groovy
duckpuppy: It was an attempt to prevent an ordering conflict between the plugins. Anything regarding a plan file is dependent only on the execution of the plan and apply commands. In my case, I wanted to skip the PLAN_COMMAND itself while still executing the PLAN stage (I need the INIT). With the above plugin bound to the entire PLAN phase, it's still going to execute even in the absence of the plan command executing and try to stash a non-existent plan file.

test/TerraformOutputOnlyPluginTest.groovy
duckpuppy: This makes sense to me. I'll update the PR with this change.

src/PassPlanFilePlugin.groovy
kmanning: Ah, I see.  Thanks for the clarification.

kmanning:
Updates look good, thanks! :shipit:

src/PassPlanFilePlugin.groovy
kmanning: Linking to Issue #152, since it touches on the concern about decorating being overly specific.

Issue #233, opened 2020-06-16
Consolidate TerraformCommand behavior


kmanning:
* `TerraformCommand` needs to be an interface, and the behavior between `TerraformValidate`/`TerraformPlan`/`TerraformApply` need  to be consolidated.  
* A lot of PR's are painful right now, without  the consolidation.

Issue #327, opened 2021-01-08
Treat default plugins like any other plugin


kmanning:
* All plugins have an `init` method to enable them - all, with the exception of "default" plugins
* Default plugins are hardcoded in specific Stages. Eg: https://github.com/manheim/terraform-pipeline/blob/master/src/TerraformEnvironmentStage.groovy#L7
* Not only does hard-coding the default plugins break the `init` convention, it makes testing a little awkward (tests need to be written in a way that assume that default plugins are present).

Let's try to reapproach default plugins in a way that makes them consistent with every other plugin.

* Let's an an `init` method to default plugins, so that they conform to the convention
* Rather than hardcoding the plugins in Stages, let's call `init` on the default plugins in `Jenkinsfile.init`.
* This may break backwards compatibility.  The order of plugins has an effect on the pipeline - anyone project that init's plugins *before* `Jenkinsfile.init` will now have a different order.  <-- Note this incompatibility when releasing this feature.

PR #349, opened 2021-03-03
Create a Pluggable trait to DRY up terraform commands

duckpuppy:
Issue #233: TerraformCommand is now a Trait, which allows for consolidating common behaviors and properties into the trait. In addition, a new Pluggable trait exists which consolidates the plugin management behaviors.


docs/Pluggable.md
```diff
@@ -0,0 +1,20 @@
...
+The trait also modifies the `toString()` method to apply the plugins prior to
+passing the method call up to the next trait in the chain, or the base class
+if it is the last trait in the chain.
+
+**NOTE** Traits are chained right to left - so the last trait in any chain is
+the first one declared in the `implements` clause.
+
+```
+class MyTerraformCommand implements TerraformCommand, Pluggable<MyTerraformCommandPlugin>
+```
```
kmanning, line 20: I think this brings up a topic of: "Docs for Users vs Docs for Developers".  This feels like a Doc for Developers, and should therefore be separated in some way from the normal docs.  I don't know what that would look like just yet.

docs/TerraformCommand.md
```diff
@@ -0,0 +1,24 @@
...
+
+The `TerraformCommand` trait overrides the default `toString()` method on the
+implementing class to return the output of `assembleCommandString()`.
+
+```
+class MyTerraformCommand implements TerraformCommand {
+  public String assembleCommandString() {
+    return terraformBinary + " help"
+  }
+}
```
kmanning, line 24: Same comment as PluggableDocs

src/Pluggable.groovy
```diff
@@ -0,0 +1,31 @@
...
+
+    public static getPlugins() {
+        return plugins
+    }
+
+    public String toString() {
+        applyPlugins()
+        // Chain the toString call to the next trait in the chain
+        return super.toString()
+    }
```
kmanning, line 30: Chaining toString() seems fraught with peril.

Conceptual speaking, toString() has nothing to do with the idea of being Pluggable.  toString is totally incidental to the way that it's being used in TerraformCommands.  In the future, I'd like to be able to extend Pluggable to other areas of terraform-pipeline, and I think toString() will hinder that future work. 

As a first-cut, I think toString() should be left out of the interface.

src/TerraformCommand.groovy
```diff
@@ -1,3 +1,26 @@
...
+    private String environment
+    private String terraformBinary = "terraform"
+
+    public String getEnvironment() {
+        return environment
+    }
+
+    public void setEnvironment(String environment) {
+        this.environment = environment
+    }
```
kmanning, line 11: Not all terraform commands have environments - eg: `TerraformValidate`.  I don't think environment should be part of the interface.

src/TerraformCommand.groovy
```diff
@@ -1,3 +1,26 @@
-interface TerraformCommand {
-    public String getEnvironment()
+trait TerraformCommand {
+    private String environment
+    private String terraformBinary = "terraform"
```
kmanning, line 3: Reaching into a trait and pulling out a private variable seems fraught with peril.  In fact, I think your change makes it clear that `terraformBinary` never belonged in TerraformCommand to begin with. It feels like this needs to be pushed out to a higher level class - I want to be able to change the terraform binary once, and have it cascade across all commands.

src/TerraformApplyCommand.groovy
```diff
@@ -95,34 +89,13 @@ class TerraformApplyCommand implements TerraformCommand, Resettable {
...
             .withArgument("-auto-approve")
     }
 
-    public static getPlugins() {
-        return plugins
-    }
-
     public static reset() {
         this.plugins = []
     }
```
kmanning, line 100: Interesting - I never thought about reset.  It seems like `reset()` could be pulled into the Pluggable interface - resetting the plugins is a fundamental part of being Pluggable.  Thoughts?

src/TerraformPlanCommand.groovy
```diff
@@ -1,19 +1,21 @@
...
     private String errorFile
     private Closure variablePattern
     private Closure mapPattern
 
+    // The static initializer is needed to initialize the static variable inherited
+    // from the Pluggable trait when the class is loaded.
+    static {
+        this.plugins = DEFAULT_PLUGINS.clone()
+    }
+
```
kmanning, line 18: Sorry about the above.  Default plugins should be handled in a different way.  If issue #327 were completed first, you wouldn't have to deal with this problem.

I might consider closing out #327 first - I don't like the idea of reaching into the private variables of a trait.

kmanning:
Looking at the extent of the changes, it feels like this may be a better change for a major release (6.0) rather than a minor (5.17)

src/TerraformCommand.groovy
duckpuppy: We're not pulling out a private variable - we're going through the public `get/set` accessors. I would typically make ALL actual variables private and get to them via `get/set`.

src/TerraformApplyCommand.groovy
duckpuppy: i was a bit confused about Resettable - I moved it around the trait hierarchy a bit and then left it out... is it JUST for plugins? It also seems like it should reset default values as well.

src/TerraformCommand.groovy
duckpuppy: I can remove it, but that does add in a bit more copy/paste.

src/TerraformPlanCommand.groovy
duckpuppy: Also `this.plugins = X` is calling the public `Pluggable.setPlugins()` method, not modifying the private variable directly.

src/Pluggable.groovy
```diff
@@ -0,0 +1,31 @@
...
+    private appliedPlugins = []
+
+    public applyPlugins() {
+        def remainingPlugins = plugins - appliedPlugins
+
+        for (T plugin in remainingPlugins) {
+            plugin.apply(this)
+            appliedPlugins << plugin
+        }
+    }
```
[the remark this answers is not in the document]
duckpuppy: As there are no other examples of things using plugins, I did work with what was there already. There's nothing really that makes it command specific though. It just calls `apply()` for each plugin that hasn't been applied yet - what `apply()` means is completely up to the implementing class.

src/TerraformApplyCommand.groovy
kmanning: No, I would not say that Resettable is JUST for plugins.  

I would say that being Pluggable automatically makes you Resettable - in so far as Plugins are stored as static variables, and Resettable is all about restoring static state as though it had never been modified (during testing).

src/Pluggable.groovy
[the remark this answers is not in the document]
kmanning: Re: "As there are no other examples of things using plugins" - oh, hrm.  Maybe I was thinking it without stating it?  Apologies.

Stages also use Plugins - eg: `TerraformEnvironmentStage`.  Stages suffer from the same problem that you've identified with TerraformCommands - there's lots of copy/paste around the adding and application of plugins.

My hope was that your carving out of Pluggable for TerraformCommands, could later be extended to address the copy/paste of the plugin code for Stages.  Hence, I love the choice of naming that you've used = `Pluggable`.  The idea of something having plugins that can be applied seems like a very generic idea, that could applicable to both TerraformCommands and to Stages.

src/Pluggable.groovy
kmanning: I guess I'll rephrase to say, I'd explicitly NOT like to include toString() to be part of contact.  And this dove tails with my comment above, about wanting to extend the idea of a `Pluggable` to Stages as well.  Applying plugins is coincidental to TerraformCommands, and would be inappropriate for a Stage.  Hence, I'd prefer to leave toString out of the contract of Pluggables to *enable* future application to Stages.

src/TerraformCommand.groovy
kmanning: Whether or not you access the variable directly, or use a get/set has very little difference.  The point is that we're reaching into the internals of the class and operating on those values externally.  That fails to encapsulate *behavior*.  

There's no *behavior* in `TerraformCommand` that relies on `terraformBinary` - hence the need to reach in, and pull it out - because the behavior lives elsewhere (the behavior does not live in the trait itself).

src/TerraformPlanCommand.groovy
kmanning: Similar to the comment above, it's less about using a get/set method, and more about reaching into a trait to pull out data and operate on that data.  

The trait `Pluggable` encapsulates behavior around plugins.  If anything, your change suggests that resetting the value of the plugins feels like a good candidate behavior to consider pushing into the trait.

kmanning: It's less about using get/set methods, and more about extracting data out of an internal class to operate on it (similar to the comment above about `terraformBinary`).

Again, this is less about what you did, and more that you've exposed an underlying problem of the library.  DefaultPlugins are treated as a special case, and they shouldn't be. If default plugins were treated like any other plugin (and added through `addPlugin`), we wouldn't be having this conversation, and you wouldn't need the `static{}` line in there.

src/Pluggable.groovy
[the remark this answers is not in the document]
duckpuppy: If I am not missing anything, the extra `applyPlugins()` applied to the toString() call is all that's making them different from the way the Stages I looked at work. I'll remove the toString() from the contract and change all the commands to not rely on toString().

duckpuppy:
What is the timing on cutting a new 6.0 release? Can that be done quickly, or are there other changes that need to be put in place before it can be done?

src/Pluggable.groovy
duckpuppy: toString() is reverted to it's previous behavior, and not part of TerraformCommand or Pluggable

src/Pluggable.groovy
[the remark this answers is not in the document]
duckpuppy: Removed toString() from Pluggable, the rest should be generic.

src/TerraformCommand.groovy
duckpuppy: Fixed.

src/TerraformCommand.groovy
duckpuppy: Removed terraformBinary from TerraformCommand

kmanning:
There's no specific timing on releases - I've been creating them on an as-needed basis.  I only carved out 6.0 for things that are large/fundamental changes, and likely to not be backwards compatible.

src/Pluggable.groovy
```diff
@@ -0,0 +1,33 @@
...
+        return plugins
+    }
+
+    public static void resetPlugins(defaultPlugins = []) {
+        this.plugins = defaultPlugins.clone()
+    }
+
+    public static void reset() {
+        this.resetPlugins()
+    }
```
kmanning, line 32: Just to confirm the expected behavior here: 
1. If I implement Pluggable, I'll automatically get a `reset()` that resets my internal plugins
2. If I have additional things that need to be reset, I can override `reset()`, lose access to `Pluggable.reset()` (by nature of the way Groovy Trait static methods work), but can still continue to reset my internal plugins by calling `resetPlugins()` which will still be accessible.

Is that correct?

docs/TerraformCommand.md
kmanning: This was reverted (meaning, TerraformCommand is still an interface, and not a trait) correct?  So it's no longer needed?

src/TerraformInitCommand.groovy
```diff
@@ -72,29 +68,11 @@ class TerraformInitCommand implements Resettable {
...
 
-    public static getPlugins() {
-        return this.globalPlugins
-    }
-
-    public static reset() {
-        globalPlugins = []
-        // This is awkward - what about the applied plugins...?
+    public String getEnvironment() {
+        return environment
```
kmanning, line 76: Did this need to be added?

test/ResetStaticStateExtension.groovy
```diff
@@ -27,6 +28,13 @@ public class ResetStaticStateExtension implements BeforeEachCallback,
...
     public findAllResettableClasses() {
-        return new Reflections(Resettable.getPackage().getName()).getSubTypesOf( Resettable.class )
+        // There's an interesting problem here.
+        def resettableClasses = new Reflections(Resettable.getPackage().getName()).getSubTypesOf( Resettable.class )
+        def traits = resettableClasses.find { Traits.isTrait(it) }
+        traits.each {
+            resettableClasses += new Reflections(it.getPackage().getName()).getSubTypesOf( it )
+            resettableClasses.removeElement(it)
+        }
+        return resettableClasses
```
kmanning, line 38: I'm not sure about the correctness of this change? 

1. Find only finds a single result, even if there are multiple
2. the `traits.each` seems to be looking for subtypes - but when I inspect the resettableClasses initially, it already contains all the subtypes - so there's no need to re-add them.

I think this could be simplified, and fixed to handle multiple traits if that problem ever cropped up in the future.

```
    return new Reflections(Resettable.getPackage().getName()).getSubTypesOf( Resettable.class ).findAll { !Traits.isTrait(it) }
```

`findAll` should select all the elements that are NOT a trait (and the only trait is Pluggable at the moment), and we're good (because all subtypes of Pluggable are already found by `getSubTypesOf(Resettable.class)`.

src/Pluggable.groovy
kmanning: Actually, I think you answer this clearly, in your change to `TerraformApplyCommand`, and `TerraformFormatCommand`.

1.  `TerraformApplyCommand` gets `reset()` "for free" from Pluggable
2. `TerraformFormatCommand` has additional things that need to be reset, and can override and do so, while maintaining plugin reset through `resetPlugins()`

duckpuppy: That is correct.

src/TerraformInitCommand.groovy
duckpuppy: No, I must have put that in by accidetnt.

kmanning:
I like what you did with moving the commends from a .md file to the classes themselves.

kmanning:
Oh, one last thing - can you update the CHANGELOG?

CHANGELOG.md
```diff
@@ -2,10 +2,10 @@
 
 * [Issue #344](https://github.com/manheim/terraform-pipeline/issues/344) Add PLAN_ONLY parameter to PlanOnlyPlugin
   * **BREAKING CHANGE** This change is a breaking change. Prior to this update, applying the PlanOnlyPlugin would restrict the pipeline to only running `terraform plan`. This update changes behavior to simply providing a `PLAN_ONLY` boolean parameter that can be set to restrict the build behavior. It defaults to `false`.
-* [Issue #347](https://github.com/manheim/terraform-pipeline/pull/348)
-  Feature: TerraformOutputOnlyPlugin - can restrict a pipeline run to
-  displaying the current state outputs only via new job parameters.
 * [Issue #318](https://github.com/manheim/terraform-pipeline/issues/318) Feature: Show environment on confirm command page
+* [Issue #347](https://github.com/manheim/terraform-pipeline/pull/348) Feature: TerraformOutputOnlyPlugin - can restrict a pipeline run to displaying the current state outputs only via new job parameters.
+* [Issue #349](https://github.com/manheim/terraform-pipeline/pull/349) Extract plugin handling to a `Pluggable` trait to DRY up the existing TerraformCommand classes.
+>>>>>>> 25c1d02 (Update CHANGELOG)
```
kmanning, line 8: Can this be cleaned up to not reorder, and to remove the `(Update CHANGELOG)` comment?

PR #350, opened 2021-03-09
Feature: Show environment on confirm apply page

### LINKED ISSUE #318: [FEATURE] - updated default confirm message to contain environment
I know we can override it but I think it would be nice if we can updated the Confirm and Delete stages to have the environment in the comment

https://github.com/manheim/terraform-pipeline/blob/f80fbe16cdf79a5511ca4b5f1f234dad24601ec6/src/ConfirmApplyPlugin.groovy#L9

`Are you absolutely sure the plan above is correct, and should be IMMEDIATELY DEPLOYED via "terraform apply" on stage ${environment}?`

thoughts?

bondezbond:
When asked to confirm the apply currently it takes you to a page with standard messaging identical for all environments. This PR gives a little more context on which environment your confirming

Closes #318


kmanning:
:shipit:

PR #353, opened 2021-03-24
Issue #354: Add taint and untaint command

### LINKED ISSUE #354: Add support for `terraform taint` and `terraform untaint`
It would be useful to have parameters that, when supplied, would trigger a `terraform taint` or `terraform untaint` prior to execution of the `terraform plan` command. This would allow forcing a replacement of a given resource, or undoing that if it's not desired.

duckpuppy:
The `TerraformTaintPlugin` adds new parameters to a build for resources to taint / untaint. If nothing is specified, then the plugin does nothing. If a value is provided in one of the parameters, then the corresponding command will be ran prior to the `terraform plan` command. It also includes support for restricting the execution to a given branch or set of branches, as well as restricting based the origin repository of the code checked out in the workspace.

```
// Jenkinsfile
@Library(['terraform-pipeline@v3.10']) _

Jenkinsfile.init(this, env)

// This enables the "taint" and "untaint" functionality
// It will only apply to the master branch (default behavior)
// and will only apply if the current worspace code is checked out from the
// MyOrg/myrepo repository.
TerraformTaintPlugin.init()
TerraformTaintPlugin.onlyOnOriginRepo("MyOrg/myrepo")

def validate = new TerraformValidateStage()

def destroyQa = new TerraformEnvironmentStage('qa')
def destroyUat = new TerraformEnvironmentStage('uat')
def destroyProd = new TerraformEnvironmentStage('prod')

validate.then(destroyQa)
        .then(destroyUat)
        .then(destroyProd)
        .build()
```

Fixes #354

### DIFF
--- a/docs/TerraformTaintPlugin.md
+++ b/docs/TerraformTaintPlugin.md
@@ -22,8 +22,6 @@ Jenkinsfile.init(this, env)
 
 // This enables the "taint" and "untaint" functionality
 // It will only apply to the master branch (default behavior)
-// and will only apply if the current worspace code is checked out from the
-// MyOrg/myrepo repository.
 TerraformTaintPlugin.init()
 
 def validate = new TerraformValidateStage()
--- a/docs/TerraformTaintPlugin.md
+++ b/docs/TerraformTaintPlugin.md
@@ -1,0 +1,46 @@
+## [TerraformTaintPlugin](../src/TerraformTaintPlugin.groovy)
+
+Enable this plugin to add `TAINT_RESOURCE` and `UNTAINT_RESOURCE` parameters
+to the build. If a resource path is provided via one of those parameters, then
+the `terraform plan` command will be preceded by the corresponding Terraform
+command to taint or untaint the appropriate resources.
+
+Note that the untaint command will take precedence, so if for some reason the
+same resource is placed in both parameters, it will be tainted and immediately
+untainted, resulting in no change.
+
+There are several ways to customize where and when the taint/untaint can run:
+
+* `onlyOnOriginRepo()`: This method takes in a string of the form
+  `<userOrOrg>/<repoName>`. When specified, the plugin will only apply when
+  the current checkout is from the same repository and not a fork or other
+  clone.
+* `onMasterOnly()`: This takes no arguments. This restricts the plugin from
+  applying to any branch other than master. This is the default behavior.
+* `onBranch()`: This takes in a branch name as a parameter. This adds the
+  branch to the list of approved branches.
+
+```
+// Jenkinsfile
+@Library(['terraform-pipeline@v3.10']) _
+
+Jenkinsfile.init(this, env)
+
+// This enables the "taint" and "untaint" functionality
+// It will only apply to the master branch (default behavior)
+// and will only apply if the current worspace code is checked out from the
+// MyOrg/myrepo repository.
+TerraformTaintPlugin.init()
+TerraformTaintPlugin.onlyOnOriginRepo("MyOrg/myrepo")
+
+def validate = new TerraformValidateStage()
+
+def destroyQa = new TerraformEnvironmentStage('qa')
+def destroyUat = new TerraformEnvironmentStage('uat')
+def destroyProd = new TerraformEnvironmentStage('prod')
+
+validate.then(destroyQa)
+        .then(destroyUat)
+        .then(destroyProd)
+        .build()
+```
--- a/docs/TerraformTaintPlugin.md
+++ b/docs/TerraformTaintPlugin.md
@@ -11,12 +11,6 @@ untainted, resulting in no change.
 
 There are several ways to customize where and when the taint/untaint can run:
 
-* `onlyOnOriginRepo()`: This method takes in a string of the form
-  `<userOrOrg>/<repoName>`. When specified, the plugin will only apply when
-  the current checkout is from the same repository and not a fork or other
-  clone.
-* `onMasterOnly()`: This takes no arguments. This restricts the plugin from
-  applying to any branch other than master. This is the default behavior.
 * `onBranch()`: This takes in a branch name as a parameter. This adds the
   branch to the list of approved branches.
 
@@ -31,7 +25,6 @@ Jenkinsfile.init(this, env)
 // and will only apply if the current worspace code is checked out from the
 // MyOrg/myrepo repository.
 TerraformTaintPlugin.init()
-TerraformTaintPlugin.onlyOnOriginRepo("MyOrg/myrepo")
 
 def validate = new TerraformValidateStage()
 
--- a/src/TerraformTaintCommand.groovy
+++ b/src/TerraformTaintCommand.groovy
@@ -1,14 +1,10 @@
-class TerraformTaintCommand implements TerraformCommand {
+class TerraformTaintCommand implements TerraformCommand, Pluggable<TerraformTaintCommandPlugin> {
   private static DEFAULT_BRANCHES = ['master']
   private static branches = DEFAULT_BRANCHES
 
-  String environment
   private String originRepoSlug = ""
-  private String terraformBinary = "terraform"
   private String command = "taint"
   private String resource
-  private static plugins = []
-  private appliedPlugins = []
 
   public TerraformTaintCommand(String environment) {
     this.environment = environment
@@ -43,9 +39,7 @@ class TerraformTaintCommand implements TerraformCommand {
     return this.environment
   }
 
-  public String toString() {
-    applyPluginsOnce()
-
+  public String assembleCommandString() {
     def parts = []
     parts << terraformBinary
     parts << command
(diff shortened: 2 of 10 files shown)


docs/TerraformTaintPlugin.md
```diff
@@ -0,0 +1,46 @@
...
+Note that the untaint command will take precedence, so if for some reason the
+same resource is placed in both parameters, it will be tainted and immediately
+untainted, resulting in no change.
+
+There are several ways to customize where and when the taint/untaint can run:
+
+* `onlyOnOriginRepo()`: This method takes in a string of the form
+  `<userOrOrg>/<repoName>`. When specified, the plugin will only apply when
+  the current checkout is from the same repository and not a fork or other
+  clone.
```
kmanning, line 17: I don't fully understand the notion of `onlyOnOriginRepo`.  If this option did not exist, how would one taint the resources from a fork? (Meaning, I'm not clear how one could taint a resource from a fork, so I'm not clear what problem this option is addressing?)

duckpuppy: The main reason this is in is because this was a requirement for the work I'm doing for this feature. Really, we don't want someone forking the project to work on it and having their builds affect state. I suppose the `onlyOnOriginRepo` could be extracted out into a separate plugin that could apply to all things.

kmanning: "Really, we don't want someone forking the project to work on it and having their builds affect state." <- Is this notion different from the notion of, "We don't want someone to run `apply` from their fork"?  

If so, the solution for that is to restrict "apply" to the master branch.  That's conceptually consistent with other plugins that affect state, and negates the need for an additional "originRepo" concept.  

I'm less concerned about extracting it to another plugin, and more concerned about introducing a concept that partially overlaps with the notion of branches.

jantman: Just wanted to weigh in a bit here...

branch != fork. I think the main concern here is for folks like me who set up their personal user as a GitHub Organization Folder in some Jenkins... and then when I fork a repo, it starts building under my Org as well as where it "should" build. So, while not being able to taint on branches other than master is probably good, it doesn't really solve the fork problem. It's really only a "problem" if someone provides parameters to the wrong job, but that's disturbingly easy to do, especially since the PR status links sent to GitHub will be links back to the fork in Jenkins.

All that being said, this might actually be a larger issue that effects pretty much everything other than "plan"... which is that you can do things on the master branch of a fork that you shouldn't be able to do anywhere other than the master branch of the upstream/canonical/origin repository...

kmanning: "this might actually be a larger issue that effects pretty much everything other than "plan"... which is that you can do things on the master branch of a fork that you shouldn't be able to do anywhere other than the master branch of the upstream/canonical/origin repository..."  <-- I agree.  If that's indeed a problem, taint/untaint doesn't seem like the appropriate place to start addressing the problem.

kmanning: I'll reprhase the above in a much more friendly way (rereading that, it sounds very unfriendly, which was not the intention).

If doing things on a fork/branch are a problem, it's indeed a big and fundamental problem.  I haven't been particularly concerned about it, because I feel like there are workarounds outside of this library that handle those concerns (ie: you can disable untrusted branches/forks at the Jenkins configuration level).  But maybe there are other concerns that you're thinking about, that haven't crossed my mind, and I'd be interested to hear them.

It seems like the problem and the likely concerns would be much broader than the narrow feature of taint/untaint, and likely deserves a separate issue, that can look at the problem across all existing plugins.

jantman: Agreed that's likely a larger issue. I'll try to flesh out #356 a bit.

kmanning:
In the meantime, @duckpuppy, how would you feel about removing the `onlyOnOriginRepo()` functionality?  If it's really needed, it could be purely additive in a fast-follow PR.  Otherwise, the discussion in #356 might bring up a better alternative.

src/TerraformTaintPlugin.groovy
```diff
@@ -0,0 +1,121 @@
...
+            apply = true
+        } else if (null == Jenkinsfile.instance.getEnv().BRANCH_NAME) {
+            apply = true
+        }
+
+        // Check repo slug
+        if (current_repo.toLowerCase() == this.origin_repo.toLowerCase()) {
+            apply = apply && true
+        } else if (! this.origin_repo) { // We assume that no origin repo means run always
+            apply = apply && true
```
kmanning, line 75: what is `apply && true` meant to accomplish?  

if `apply` is true, then `true && true` = `true`
if `apply` is false, then `false && true` = `false`

apply isn't being altered in either case?

src/TerraformTaintPlugin.groovy
```diff
@@ -0,0 +1,121 @@
...
+        if (! originUrl.endsWith('.git')) { originUrl = "${originUrl}.git" }
+        def giturl = originUrl =~ /^.*@.*[:\/]([^\/]+\/.*)\.git$/
+        def httpurl = originUrl =~ /^https?:\/\/[^\/]+\/([^\/]+\/[^\/]+)\.git$/
+        if ( giturl.matches() ) {
+            return giturl[0][1]
+        } else if ( httpurl.matches() ) {
+            return httpurl[0][1]
+        }
+        return ""
+    }
```
kmanning, line 93: Can the method in `Jenkinsfile` work here?  https://github.com/manheim/terraform-pipeline/blob/master/src/Jenkinsfile.groovy#L8-L17

src/TerraformTaintPlugin.groovy
```diff
@@ -0,0 +1,121 @@
...
+
+    public static onlyOnOriginRepo(String origin_repo) {
+        this.origin_repo = origin_repo
+        return this
+    }
+
+    public static onMasterOnly() {
+        this.branches = ['master']
+        return this
+    }
```
kmanning, line 34: Is `onMasterOnly` needed, if this is the default?   In what case would a user need to use `onMasterOnly`?

docs/TerraformTaintPlugin.md
```diff
@@ -0,0 +1,39 @@
...
+```
+// Jenkinsfile
+@Library(['terraform-pipeline@v3.10']) _
+
+Jenkinsfile.init(this, env)
+
+// This enables the "taint" and "untaint" functionality
+// It will only apply to the master branch (default behavior)
+// and will only apply if the current worspace code is checked out from the
+// MyOrg/myrepo repository.
```
kmanning, line 26: This statement is no longer needed, right? 
```
// and will only apply if the current worspace code is checked out from the
// MyOrg/myrepo repository.
```

test/TerraformTaintCommandTest.groovy
```diff
@@ -0,0 +1,77 @@
...
+@ExtendWith(ResetStaticStateExtension.class)
+class TerraformTaintCommandTest {
+    @Nested
+    public class WithResource {
+        @Test
+        void defaultsToEmpty() {
+            def command = new TerraformTaintCommand()
+
+            def actualCommand = command.toString()
+            assertThat(actualCommand, equalTo("echo \"No resource set, skipping 'terraform taint'."))
```
kmanning, line 20: I'm fine with this, but looking at the source code, this isn't actually going to execute is it?  `command.toString()` is called within a guard that should ensure that the resource was set.  

```
            if (untaintCommand.resource && shouldApply()) {
                echo "Running '${untaintCommand.toString()}'. TerraformTaintPlugin is enabled."
                sh untaintCommand.toString()
            }
```

I'd be fine with taking this functionality out ("this functionality" = echo if resource is not set), and letting the command fail on its own (ie: it should fail because `terraform taint` expects a resource which will not be provided, and will return a non-zero response code and display the help command).  But I'm fine leaving it in if you think it's a better experience.

test/TerraformUntaintCommandTest.groovy
```diff
@@ -0,0 +1,77 @@
...
+class TerraformUntaintCommandTest {
+    @Nested
+    public class WithResource {
+        @Test
+        void defaultsToEmpty() {
+            def command = new TerraformUntaintCommand()
+
+            def actualCommand = command.toString()
+            assertThat(actualCommand, equalTo("echo \"No resource set, skipping 'terraform untaint'."))
+        }
```
kmanning, line 21: Same comment as above - will this actually ever execute in a pipeline?

test/TerraformTaintCommandTest.groovy
duckpuppy: It's true that the test code is testing code that won't execute in practice. I'm not sure where to fall on failure. I don't mind it failing by having the command blow up, it's more obvious to the user.

I'll remove the guard on having the resource set.

src/TerraformTaintCommand.groovy
```diff
@@ -0,0 +1,37 @@
+class TerraformTaintCommand implements TerraformCommand, Pluggable<TerraformTaintCommandPlugin> {
+    private String command = "taint"
+    private String resource
+    private String environment
```
kmanning, line 4: Is `environment` dead code?

kmanning: Gross, it's part of the `TerraformCommand` interface.

kmanning: Cutting a separate issue for that - https://github.com/manheim/terraform-pipeline/issues/357

kmanning:
Looks good, :shipit:

src/TerraformTaintCommand.groovy
duckpuppy: Yeah, that was part of why I moved environment management into it when I was converting it to a trait - it already had getEnvronment() in the interface.

Issue #359, opened 2021-04-05
Master branch fails to run successfully in Jenkins


kmanning:
Sample Pipeline: 
```
@Library(['terraform-pipeline-manheim@master') _

Jenkinsfile.init(this)

def validate = new TerraformValidateStage()
def deployQa = new TerraformEnvironmentStage('qa')
def deployUat = new TerraformEnvironmentStage('uat')

validate.then(deployQa)
        .then(deployUat)
        .build()
```

Expected: Pipeline works
Actual: 
```
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
file:/var/jenkins_home/jobs/terraform/jobs/fakeDb/branches/docker/builds/8/libs/terraform-pipeline-keith/src/TerraformFormatCommand.groovy: -1: Unsupported array expression for CPS transformation in this context @ line -1, column -1.
file:/var/jenkins_home/jobs/terraform/jobs/fakeDb/branches/docker/builds/8/libs/terraform-pipeline-keith/src/TerraformInitCommand.groovy: -1: Unsupported array expression for CPS transformation in this context @ line -1, column -1.
file:/var/jenkins_home/jobs/terraform/jobs/fakeDb/branches/docker/builds/8/libs/terraform-pipeline-keith/src/TerraformPlanCommand.groovy: -1: Unsupported array expression for CPS transformation in this context @ line -1, column -1.
file:/var/jenkins_home/jobs/terraform/jobs/fakeDb/branches/docker/builds/8/libs/terraform-pipeline-keith/src/TerraformApplyCommand.groovy: -1: Unsupported array expression for CPS transformation in this context @ line -1, column -1.
file:/var/jenkins_home/jobs/terraform/jobs/fakeDb/branches/docker/builds/8/libs/terraform-pipeline-keith/src/TerraformValidateCommand.groovy: -1: Unsupported array expression for CPS transformation in this context @ line -1, column -1.
```

Seems to be related to merge of https://github.com/manheim/terraform-pipeline/pull/353.  Looks like CPS may not support the Trait refactoring that was done as part of the merge.

PR #360, opened 2021-04-05
Fix CPS issues

duckpuppy:
In order to remove CPS transform issues, the Pluggable interface has been removed and plugin management is now back in the individual Terraform*Command classes. I will continue to investigate other methods of DRYing up things like plugins.


kmanning:
Linking this to Issue #359

kmanning:
Looks good, thanks for the fix.  :shipit:

Issue #368, opened 2021-04-16
FlywayPlugin: Prompt the user *again* before applying migration


kmanning:
* If a pending migration is detected, when the user hits apply, prompt the user *again*, with a message, "There is a pending migration - are you sure you want to apply it?"
* Migrations will likely be an infrequent event, so this is a way to optionally draw extra attention to a developer, to make them aware that they're making a change that they might not realize.
* If this feature is not desired, the user can optionally disable it.

PR #371, opened 2021-04-20
Issue 368: Prompt the user *again* if a pending migration is detected.  Optionally disable.

kmanning:
* Issue #368 
* Applying a migration may be an uncommon case - more commonly, no pending migration may exist
* Because this is an uncommon occurrence, a user may not be accustomed to validating during the "plan" stage whether or not a pending migration exists.
* As an extra safe guard, optionally prompt the user *again* if a migration is detected - reconfirm that the user is aware that a pending migration exists, and approves the application of said migration.
* This should be optional


kmanning:
Suggested implementation: 
* Tee the flyway output (and handle failure cases properly)
* Grep the output for `Pending` (The various statuses are available here - https://flywaydb.org/documentation/concepts/migrations#migration-states)
* If a `Pending` status is detected, prompt the user, "A pending migration was found - please review your flyway output.  Do you want to continue and apply the migration?"

PR #375, opened 2021-04-27
Fix taint and untaint behavior

duckpuppy:
This PR fixes a bug with the TerraformTaintPlugin where the taint or untaint commands would run even if no resources were specified.


kmanning:
Issue #354

kmanning (approved):
:shipit:

Issue #309, opened 2020-11-10
Add the ability to specify terraform-pipeline starting execution workspace directory


vincentclee:
Currently, we have the **TerraformDirectoryPlugin** which allows **terraform-pipeline** to do this:

```
TerraformDirectoryPlugin.withDirectory("app1/terraform").init()
```
```
terraform <command> app1/terraform
...
```

What i would like **terraform-pipeline** to support:

```
TerraformStartingDirectoryPlugin.withDirectory("appQ/terraform").init()

```
```
cd appQ/terraform
terraform <command> .
```

##

This is important with monolithic pipeline repo's where the `.terraform-version` flat file is only read from the root of the project tree when using the **TerraformDirectoryPlugin**.

kmanning:
Seems like a good UseCase.  In thinking through this, we'll have to be careful about how other TerraformCommandPlugins interact with this new UseCase.  Lots of plugins decorate the front of the command (eg: awssume/parameter-store-exec), and the command itself is treated as a single shell step.  What you're looking for is a separate shell step.

kmanning:
I don't know that TerraformDirectoryPlugin is so inappropriately named for this UseCase.  I wonder if it could offer both behaviors with the same plugin? 

```
TerraformDirectoryPlugin.withDirectory("app1/terraform").init()
```
vs 
```
TerraformDirectoryPlugin.withStartDirectory("app1/terraform").init()
```
or something like
```
TerraformDirectoryPlugin.withDirectory("app1/terraform").withCdCommand().init()
```

vincentclee:
I like option one.
Extend **TerraformDirectoryPlugin** to support the `withStartDirectory` method

kmanning:
Note: Change in terraform 0.14 and 0.15 related to subdirectories.  https://www.terraform.io/docs/cli/commands/plan.html#passing-a-different-configuration-directory

vincentclee:
bump

vincentclee:
This is officially needed on terraform 0.15 and later.

kdibenedetto:
Yeah, I went to use tf 1.0.1 and was hit with this issue.

PR #398, opened 2021-07-28
TerraformStartDirectoryPlugin - Change the starting execution workspace directory

vincentclee:
* https://github.com/manheim/terraform-pipeline/issues/309

### DIFF
--- a/src/TerraformStartDirectoryPlugin.groovy
+++ b/src/TerraformStartDirectoryPlugin.groovy
@@ -1,0 +1,42 @@
+import static TerraformValidateStage.ALL
+import static TerraformEnvironmentStage.ALL
+
+public class TerraformStartDirectoryPlugin implements TerraformValidateStagePlugin, TerraformEnvironmentStagePlugin, Resettable {
+    private static String directory = "./terraform/"
+
+    TerraformStartDirectoryPlugin() { }
+
+    public static void init() {
+        def plugin = new TerraformStartDirectoryPlugin()
+
+        TerraformValidateStage.addPlugin(plugin)
+        TerraformEnvironmentStage.addPlugin(plugin)
+    }
+
+    public static withDirectory(String directory) {
+        this.directory = directory
+        return this
+    }
+
+    @Override
+    public void apply(TerraformValidateStage stage) {
+        stage.decorate(TerraformValidateStage.ALL, addDirectory())
+    }
+
+    @Override
+    public void apply(TerraformEnvironmentStage stage) {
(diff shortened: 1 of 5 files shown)


kmanning:
Thanks for the contribution!  It's looking good so far.

No rush if this is a WorkInProgress, but before you're ready to merge, the coverage reports show a lack of coverage in 2 methods - `withDirectory`, and `addDirectory`.  

Typically for configuration methods like `withDirectory`, I simply validate that the method is fluent (that it can be chained with other methods, by returning itself).  Eg: `FlywayMigrationPlugin.withEchoEnabled()` - https://github.com/manheim/terraform-pipeline/blob/95077845e4324d4ba45850fbd3c573ff97fe8a6a/test/FlywayMigrationPluginTest.groovy#L71-L79

For methods like `addDirectory`, it's exercising Jenkinsfile DSL methods.  `MockWorkflowScript` is available to make testing those methods easier.  There usually 2 general things I test for:

1. if it's meant to wrap around other code, it's *REALLY* important that the wrapped code actually executes.  A failure to execute the wrapped code means the Stages basically do nothing.  So for all Jenkinsfile DSL closure methods like `addDirectory`, you'll typically see a `runsTheNestedClosure` to guarantee that a problem like that doesn't slip through.  https://github.com/manheim/terraform-pipeline/blob/95077845e4324d4ba45850fbd3c573ff97fe8a6a/test/FlywayMigrationPluginTest.groovy#L84-L94
2. it does the thing it was intended to do.  In your case, call the Jenkinsfile DSL `dir` method.  That's can be accomplished by a `verify` assertion on `MockWorkflowScript`.  Eg: https://github.com/manheim/terraform-pipeline/blob/95077845e4324d4ba45850fbd3c573ff97fe8a6a/test/FlywayMigrationPluginTest.groovy#L96-L108

src/TerraformStartDirectoryPlugin.groovy
```diff
@@ -0,0 +1,42 @@
+import static TerraformValidateStage.ALL
+import static TerraformEnvironmentStage.ALL
+
+public class TerraformStartDirectoryPlugin implements TerraformValidateStagePlugin, TerraformEnvironmentStagePlugin, Resettable {
+    private static String directory = "./terraform/"
+
+    TerraformStartDirectoryPlugin() { }
```
kmanning, line 7: FYI: Once you define a class, Groovy will give you a default constructor method for free.  So while there's nothing wrong with defining it here, you're not required to provide it.

kmanning:
Looks great, thanks for the contribution!

PR #408, opened 2021-09-14
support http/2 client response headers for GithubPRPlugin check

### LINKED ISSUE #407: HTTP/2 201 fails PR Posting
If the response uses http/2 the jobs fail to move past even though it was successful in posting the comment back to the PR. 

https://github.com/manheim/terraform-pipeline/blob/a14c20db5eca4587d025de4e3ca7b093700819a5/src/GithubPRPlanPlugin.groovy#L182

Fix should be to check for `HTTP/2 201` as well

codezninja:
## Merge Checklist

* [X] Have you linked this Pull Request to an [Issue](https://github.com/manheim/terraform-pipeline/issues)?
* [ ] Have you updated any relevant README files, to explain how this new feature works (with examples)?
* [X] Have you added relevant test cases for your change?
* [X] Do all tests and code style checks pass with `./gradlew check --info`?
* [ ] Have you successfully run your change in a pipeline in a Jenkins instance?

Fixes #407

### DIFF
--- a/src/GithubPRPlanPlugin.groovy
+++ b/src/GithubPRPlanPlugin.groovy
@@ -179,7 +179,7 @@ class GithubPRPlanPlugin implements TerraformPlanCommandPlugin, TerraformEnviron
                 def output = sh(script: cmd, returnStdout: true).trim()
 
                 def headers = readFile('comment.headers').trim()
-                if (! headers.contains('HTTP/1.1 201 Created')) {
+                if (! (headers.contains('HTTP/1.1 201 Created') || headers.contains('HTTP/2 201 Created'))) {
                     error("Creating GitHub comment failed: ${headers}\n")
                 }
                 // ok, success
--- a/src/GithubPRPlanPlugin.groovy
+++ b/src/GithubPRPlanPlugin.groovy
@@ -179,7 +179,7 @@ class GithubPRPlanPlugin implements TerraformPlanCommandPlugin, TerraformEnviron
                 def output = sh(script: cmd, returnStdout: true).trim()
 
                 def headers = readFile('comment.headers').trim()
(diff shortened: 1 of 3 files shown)


src/GithubPRPlanPlugin.groovy
```diff
@@ -179,7 +179,7 @@ class GithubPRPlanPlugin implements TerraformPlanCommandPlugin, TerraformEnviron
                 def output = sh(script: cmd, returnStdout: true).trim()
 
                 def headers = readFile('comment.headers').trim()
-                if (! headers.contains('HTTP/1.1 201 Created')) {
+                if (! (headers.contains('HTTP/1.1 201 Created') || headers.contains('HTTP/2 201 Created'))) {
```
kmanning, line 182: Do we really care about the HTTP version?  Could we simplify this to "201 Created"?

Not a change that you need to make, but arguably in the future, we could even git rid of the comment.headers, with a 

`[ curl -o output -I -L -s -w "%{http_code}" "${pullRequestUrl}" -eq 201 ]`

bondezbond:
need to merge #410 before we can fix this

src/GithubPRPlanPlugin.groovy
bondezbond: I agree we can definitely simplify and check that. @jleopold28 thoughts?

jleopold28: @bondezbond checking response code sounds good to me 👍

bondezbond: done

bondezbond: @kmanning during testing a colleague saw that http/2 only returns `201` now and drops `Created` part in response. So for now we'll have to to keep that complex conditional or we can move to a simplified curl call during this PR.

kmanning (approved):
:shipit:

Issue #379, opened 2021-04-30
Support terraform 0.15


kmanning:
* This is a placeholder.  Please note any changes that need to be made to support terraform 15

PR #415, opened 2022-01-31
Add chdir argument to support Terraform 0.15+

jleopold28:
## Merge Checklist

Resolves Issue https://github.com/manheim/terraform-pipeline/issues/379 and https://github.com/manheim/terraform-pipeline/issues/395

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,6 @@
 # Unreleased
 - [Issue #379](https://github.com/manheim/terraform-pipeline/issues/379) Support Terraform 0.15. Added `-chdir` argument.
 - [Issue #395](https://github.com/manheim/terraform-pipeline/issues/395) Support Terraform 1.0. Added `-chdir` argument.
-- Modified ``PassPlanFilePlugin``
-  - Additional argument ``withDirectory``. This allows stash and unstash commands to be run in subdirectory.
 
 
 # v5.18
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
-- [Issue #379](https://github.com/manheim/terraform-pipeline/issues/379) Support Terraform 0.15+. Added `-chdir` argument.
+- [Issue #379](https://github.com/manheim/terraform-pipeline/issues/379) Support Terraform 0.15. Added `-chdir` argument.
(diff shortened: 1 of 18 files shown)


kmanning:
Will this break versions prior to 0.15?  The library is designed to support multiple different versions of terraform simultaneously, and I'd like to preserve that if possible. 

That that end, TerraformPlugin and the TerraformPluginVersion were designed to encapsulate version-specific differences.  Could this be a means for implementing the `-chdir` flag for newer versions, while maintaining existing functionality for existing versions?

jleopold28:
@kmanning this will break versions to 0.15 ONLY when using the TerraformDirectoryPlugin. I'm going to look at making this backwards compatible

kmanning:
Good work, and nice catches on the PassPlanPlugin, and TerraformVersion dependencies.

This PR is starting to get a little big, and I'm seeing 3 logically separate pieces: 
1. TerraformVersion was missing a hook for TerraformInit
2. TerraformDirectoryPlugin can be used independently of PassPlanPlugin.  In terms of PRs, improving TerraformDirectoryPlugin would provide value for users that do not use PassPlanPlugin.
3. Update (and fix the naming conventions) for PassPlanPlugin.

That might help in terms of review.

=====

For the PassPlan change, I'll repost these on that specific PR (but I'll note them here so you know ahead of time):

1. if an Issue doesn't already exist, could you create one?  I'd like to ensure that the CHANGELOG relates all changes to a relevant Issue.  See: https://github.com/manheim/terraform-pipeline#how-to-contribute
2. Could the README be split into 2 examples?  The first example should be as-is - just the PassPlanPlugin.  Then your change could introduce a second example of how it can be used in conjunction with TerraformDirectory.  To the extent that is possible, I'd like the first example of every plugin to be the minimum code necessary to get value out of the plugin.  Every subsequent example should illustrate variations to show how to get more value out.
3. [PassPlanFilePluginTest.groovy](https://github.com/manheim/terraform-pipeline/pull/415/files#diff-47e76b1bbf42ddb9b775101f47b1223788776d0c78a74f43300f4875d656ece6)  The apply method doesn't actually care about whether or not a directory is specified.  Is the stash/unstash closures that care.  I think the tests might be more focused if they were written against those methods, rather than the `apply` method.

kmanning (approved):
Looks great - :shipit:

Issue #420, opened 2022-02-03
PassPlanFilePlugin: Add withDirectory argument


jleopold28:
Additional argument ``withDirectory`` should be added to PassPlanFilePlugin. This allows stash and unstash commands to be run in subdirectory.

PR #421, opened 2022-02-03
Support withDirectory for PassPlanFilePlugin

jleopold28:
## Merge Checklist

Resolves issue https://github.com/manheim/terraform-pipeline/issues/420

### DIFF
--- a/test/PassPlanFilePluginTest.groovy
+++ b/test/PassPlanFilePluginTest.groovy
@@ -84,13 +84,13 @@ class PassPlanFilePluginTest {
         @Test
         void runsStashPlan() {
             def plugin = new PassPlanFilePlugin()
-            def workflowScript = mock(MockWorkflowScript.class)
+            def workflowScript = spy(new MockWorkflowScript())
 
             def stashClosure = plugin.stashPlan('dev')
             stashClosure.delegate = workflowScript
             stashClosure.call { } // we don't care about the inner closure, so we're passing an empty one
 
-            verify(workflowScript, times(1)).stash(eq({name: 'tfplan-dev', includes: 'tfplan-dev'}))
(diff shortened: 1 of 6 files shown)


test/PassPlanFilePluginTest.groovy
```diff
@@ -45,35 +45,73 @@ class PassPlanFilePluginTest {
...
+
+            @Test
+            void addsArgumentToTerraformApply() {
+                PassPlanFilePlugin plugin = new PassPlanFilePlugin()
+                TerraformApplyCommand command = new TerraformApplyCommand("dev")
+                plugin.apply(command)
+
+                String result = command.toString()
+                assertThat(result, containsString("tfplan-dev"))
+            }
```
kmanning, line 114: `apply` hasn't changed at all with this PR, just `stash`/`unstash`.  It feels like it would make more sense to apply the tests to those contexts, and leave `apply` as-is. 

In the case of testing `stash`/`unstash`, you could validate that by default, `MockWorkflow.dir` is given './', and then when you call `PassPlanFilePlugin.withDirectory('blah')`, `MockWorkflow.dir` is given 'blah'

kmanning:
Also, it may make sense to validate that `withDirectory` is fluent.

kmanning:
Also, this PR introduces static state into the PassPlanPlugin for the very first time.  You've added the `reset()` method that can be used to reset state during unit testing, but there's no way for it to be called yet.  Can you have the class inherit the `Resettable` interface?  https://github.com/manheim/terraform-pipeline/blob/master/src/Resettable.groovy

The PassPlanTest class is using the ResetStaticStateExtension, but it operates by calling `reset` on all classes that implement Resettable - [ResetStaticStateExtension](https://github.com/manheim/terraform-pipeline/blob/master/test/ResetStaticStateExtension.groovy#L31)

kmanning:
Looks great - :shipit:

kmanning (approved):

PR #433, opened 2022-04-13
[Bugfix] Fixes #417 - Force terraform command to `apply` if a plan file is being used

### LINKED ISSUE #417: DestroyPlugin & PassPlanFilePlugin - Terraform Destroy can't be called with a plan file
The `DestroyPlugin` does not play nicely with `PassPlanFilePlugin`.

**Error Message**

```
 + terraform destroy -input=false -auto-approve tfplan-uat
 ╷
 │ Error: Destroy can't be called with a plan file
 │ 
 │ If this plan was created using plan -destroy, apply it using:
 │   terraform apply "tfplan-uat"
 ╵
```

**Jenkinsfile**
```groovy
Jenkinsfile.init(this)
PassPlanFilePlugin.init()
DestroyPlugin.init()

def validate = new TerraformValidateStage()
def destroyUat = new TerraformEnvironmentStage("uat")

validate.then(destroyUat).build()
```

**Terraform Command: destroy**
> The terraform destroy command accepts most of the options that `terraform apply` accepts, although it does not accept a plan file argument and forces the selection of the "destroy" planning mode.

https://www.terraform.io/cli/commands/destroy

duckpuppy:
If a plan file is being used in the `TerraformApplyCommand`, then the only
applicable Terraform subcommand is `apply`. This PR forces apply to be used if
a plan file is set. This works whether the plan file was created with or without
the `-destroy` argument.

- Force the use of `terraform apply` in TerraformApplyCommand if a planFile is present
- Update changelog

## Merge Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
- - [Issue #432](https://github.com/manheim/terraform-pipeline/issues/432) pass TagPlugin through `-var-file={env}-tags.tfvars`
+- [Issue #432](https://github.com/manheim/terraform-pipeline/issues/432) pass TagPlugin through `-var-file={env}-tags.tfvars`
+- [Issue #417](https://github.com/manheim/terraform-pipeline/issues/417) DestroyPlugin & PassPlanFilePlugin - Terraform Destroy can't be called with a plan file
 
 # v5.19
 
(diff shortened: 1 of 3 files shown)


kmanning:
1. Is there an open issue for this?  
2. In the Issue, can you specify the version of terraform that you're using, along with the behavior you're expecting, and the behavior you're seeing? 
3.  It looks like in terraform 0.15.2+, the `-destroy` flag was added to `terraform apply`.  So for later versions, it seems like the correct fix would be to change the DestoryPlugin to pass the flag, instead of altering the command.
4. Separate from the version question - are the 2 plugins compatible?  Why would you not simply remove/disable the plan plugin?

kmanning (approved):
:shipit:

PR #440, opened 2022-07-07
Fix potential infinite recursion in TerraformImportCommand

### LINKED ISSUE #439: TerraformImportCommand may enter an infinite loop
When using the TerraformImportPlugin, you may see an error similar to the following:

```
java.lang.StackOverflowError: Excessively nested closures/functions at TerraformImportCommand.getResource(TerraformImportCommand.groovy:46) - look for unbounded recursion - call depth: 1025
```

The code in question is a simple method
```groovy
    public String getResource() {
        return this.resource
    }
```

It's possible that the `this.` indirection in the return statement is causing a recursive call to `getResouce()` in some cases.

duckpuppy:
Fixes #439

## Merge Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@ build
 .idea
 *.iml
 /out
+.history
+bin
+.vscode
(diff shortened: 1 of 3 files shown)


kmanning:
There's documentation for this feature - https://github.com/manheim/terraform-pipeline/blob/master/docs/TerraformImportPlugin.md

1. Is the documentation correct?
2. Does the example from the documentation reproduce the error?
3. Were you able to reproduce the error? (Ie: do we know that this will fix the problem)

PR #445, opened 2022-11-01
Fixes #444 - Add duration parameter to WithAwsPlugin.withRole()

### LINKED ISSUE #444: WithAwsPlugin - Add configurable duration (or other options) for longer sessions
The [WithAwsPlugin](https://github.com/manheim/terraform-pipeline/blob/master/src/WithAwsPlugin.groovy) currently only supports specifying the `role` (ARN) parameter for the underlying [withAws](https://www.jenkins.io/doc/pipeline/steps/pipeline-aws/#withaws-set-aws-settings-for-nested-block) pipeline step.

This results in credentials that are only valid for 1 hour, and problematic issues with long-running operations like creating RDS read instances... at worst, if using S3 remote state, a single operation that takes longer than an hour results in the credentials expiring and not being able to update the state file. This means the state file is out of sync, and may result in orphaned resources.

The simple fix here would be to allow `withRole` to take an optional integer duration parameter which is passed on to `withAws()`. Even better would be to allow `withRole` to accept any additional parameters (e.g. `externalId`, `roleSessionName`, etc.) that `withAws()` accepts, and pass them through to the underlying `withAws()` call.

(Opening this issue for a colleague, I didn't experience it myself.)

jantman:
Work in progress PR to fix #444 by adding an optional `duration` parameter to WithAwsPlugin.withRole().

## Merge Checklist

### DIFF
--- a/docs/WithAwsPlugin.md
+++ b/docs/WithAwsPlugin.md
@@ -61,8 +61,14 @@ validate.then(deployQa)
         .build()
 ```
 
-If you want to specify a role session duration other than the default of 1 hour (3600 seconds), you can do so by providing an integer `duration` parameter to the `withRole()` call, like:
+If you want to specify a role session duration other than the default of 1 hour (3600 seconds), you can do so by providing an integer `duration` parameter to the `withRole()` call. You _must_ specify the Role ARN in this case, even if `null`. i.e.:
 
 ```
-WithAwsPlugin.withRole(duration: 43200).init()
+WithAwsPlugin.withRole('MY_ROLE_ARN', 43200).init()
+```
+
+or, to use the default environment variables for the role ARN:
+
+```
+WithAwsPlugin.withRole(null, 43200).init()
 ```
--- a/docs/WithAwsPlugin.md
+++ b/docs/WithAwsPlugin.md
@@ -60,3 +60,9 @@ validate.then(deployQa)
         .then(deployProd)
         .build()
 ```
+
+If you want to specify a role session duration other than the default of 1 hour (3600 seconds), you can do so by providing an integer `duration` parameter to the `withRole()` call, like:
+
+```
+WithAwsPlugin.withRole(duration: 43200).init()
+```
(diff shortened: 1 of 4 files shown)


bgshacklett:
This is working in at least one pipeline, now!

docs/WithAwsPlugin.md
```diff
@@ -60,3 +60,15 @@ validate.then(deployQa)
...
+
+```
+WithAwsPlugin.withRole('MY_ROLE_ARN', 43200).init()
+```
+
+or, to use the default environment variables for the role ARN:
+
+```
+WithAwsPlugin.withRole(null, 43200).init()
+```
```
kmanning, line 74: This fix looks good overall, but one minor tweak.  

This feature is an example of trying to configure a plugin that has multiple different parameters, which can be set independently of one another. 

Rather than trying to overload the `.withRole()` method to optionally accept a role/role+duration/duration, can we simply add a `.withDuration()` method?  That way, if you want to set duration, you can do so explicitly.  If you want to set the role, you can do so explicitly.  If you want to set them both, then you chain and call them both explicitly.

We can avoid the awkardness and lack-of-context in your example above: `WithAwsPlugin.withRole(null, 43200)`.  "What is 43200?  What does that have to do with a null role?"  vs `WithAwsPlugin.withDuration(43200)` which gives context to 43200.  "43200 is probably a duration".

This is a convention that's used heavily across all the plugins, so I'd like to keep that consistent.  https://en.wikipedia.org/wiki/Fluent_interface

jantman:
Rebased on master, and finally getting around to making the changes requested in the previous comment.

jantman:
@kmanning I've (finally) updated this PR as per the discussion above. Note that while the unit tests pass, I have _not_ tested this in an actual pipeline yet, as I don't have one appropriate for that.

@bgshacklett Is there any chance that, when you have time, you could update your pipeline with these changes?

kmanning (approved):
:shipit:

vincentclee:
I am continuing where @bgshacklett left off. This change has been tested internally and is ready to be merged and a release cut. Thank You.

